PR #1, opened 2020-09-21
Integrated firebase into the app

jrosen081:
Firebase will be hidden behind the interactor, so only the interactor will be used.


cindy1u0 (approved):
Looks good!!

PR #2, opened 2020-09-22
Integrated firebase cloud storage

jrosen081:
image
👑

### DIFF
--- a/frontend/src/data/reducer.js
+++ b/frontend/src/data/reducer.js
@@ -2,6 +2,7 @@ import { types } from "./actions";
 
 const initialState = {
   id: null,
+  imageURL: null,
 };
 
 const reducer = (state = initialState, action) => {
@@ -10,6 +11,7 @@ const reducer = (state = initialState, action) => {
     case types.SUCCESS:
       return {
         id: state.id || payload.id,
+        imageURL: state.imageURL || payload.imageURL,
       };
     case types.REQUEST:
     case types.ERROR:
--- a/frontend/src/data/saga.js
+++ b/frontend/src/data/saga.js
@@ -11,6 +11,7 @@ export default function* rootSaga() {
 function* root() {
   yield takeLatest(types.REQUEST, watchSingleRequest);
   yield takeLatest(types.ADDUSER, watchAddUser);
+  yield takeLatest(types.DOWNLOADIMAGE, watchDownloadImage);
 }
 
 function* watchSingleRequest() {
@@ -33,6 +34,21 @@ function* watchAddUser(action) {
     yield call(updateWithSuccess);
     yield put(singleRequest.success({ id: id }));
   } catch (error) {
+    yield put(singleRequest.error());
+  }
+}
+
+function* watchDownloadImage(action) {
+  let { imageURL } = action.payload;
+  try {
+    let image;
+    const updateWithSuccess = async () => {
+      image = await firebaseInteractor.downloadImage(imageURL);
+    };
+    yield call(updateWithSuccess);
+    yield put(singleRequest.success({ imageURL: image }));
+  } catch (error) {
+    console.log(error);
     yield put(singleRequest.error());
   }
 }
--- a/frontend/src/components/Layout.js
+++ b/frontend/src/components/Layout.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import { connect } from "react-redux";
 import styled from "styled-components";
 import { singleRequest } from "../data/actions";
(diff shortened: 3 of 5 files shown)


frontend/src/data/reducer.js
```diff
@@ -10,6 +11,7 @@ const reducer = (state = initialState, action) => {
     case types.SUCCESS:
       return {
         id: state.id || payload.id,
+        imageURL: state.imageURL || payload.imageURL,
```
cindy1u0, line 14: I think it's okay to just do imageURL: payload.imageURL because if it's not successfully, it will go back to the original state. Same thing with the id part

frontend/src/data/saga.js
```diff
@@ -25,14 +26,40 @@ function* watchSingleRequest() {
...
+}
+
+function* watchDownloadImage(action) {
+  yield performAsyncOperation(
+    () => firebaseInteractor.downloadImage(action.payload.imageURL),
+    "imageURL"
+  );
+}
+
+/**
```
cindy1u0: OmO templates

frontend/src/data/saga.js
```diff
@@ -25,14 +26,40 @@ function* watchSingleRequest() {
...
+  let value;
+  const updateWithSuccess = async () => {
+    value = await operation();
+  };
   try {
-    let id;
-    const updateWithSuccess = async () => {
-      id = await firebaseInteractor.addUser(name);
-    };
     yield call(updateWithSuccess);
```
cindy1u0, line 48: I wonder if you can just do yield call(operation) because yield call is already doing asynchronous call

frontend/src/data/saga.js
```diff
@@ -25,14 +26,40 @@ function* watchSingleRequest() {
 
 function* watchAddUser(action) {
   let { name } = action.payload;
+  yield performAsyncOperation(() => firebaseInteractor.addUser(name), "id");
+}
+
+function* watchDownloadImage(action) {
```
cindy1u0, line 41: Hmm i wonder if you can combine watchDownloadImage with performAsyncOperation

frontend/src/data/saga.js
```diff
@@ -25,14 +26,40 @@ function* watchSingleRequest() {
...
-    let id;
-    const updateWithSuccess = async () => {
-      id = await firebaseInteractor.addUser(name);
-    };
     yield call(updateWithSuccess);
-    yield put(singleRequest.success({ id: id }));
+    // Create the object to return, using the payloadName and the returned value
+    let returnObject = {};
+    returnObject[payloadName] = value;
+    yield put(singleRequest.success(returnObject));
```
cindy1u0: I think this can just be singleRequest.success({ imageURL: value })

cindy1u0:
The image is beautiful HAHAHAHA

frontend/src/data/reducer.js
jrosen081: The  reason I did it like this is that there  are now two state updates (id and imageURL). If one was already set, then I don't want it to be overridden with undefined in the other state update. Does that seem right?

frontend/src/data/saga.js
jrosen081: The reason I added the `updateWithSuccess` function is that I need to actually get the value from the response to send to the success function.

frontend/src/data/saga.js
jrosen081: `performAsyncOperation` is what actually does the async. I was thinking that each of these functions that watch will call the `performAsyncOperation` when applicable

frontend/src/data/saga.js
jrosen081: This is in the `performAsyncOperation`, which is abstract over the payloadName. I wonder if the diff looks weird, I didn't actually check it.

frontend/src/data/reducer.js
cindy1u0: Ah I see. I think it will still keep the previous updated state cuz it depends on what you were updating. I haven't tried it tho

frontend/src/data/saga.js
cindy1u0: ah gotcha

frontend/src/data/saga.js
cindy1u0: Oh interesting ill try to run it too

cindy1u0 (approved):

frontend/src/data/saga.js
cindy1u0: I would say stay consistent? Cuz like in our success in action, it's doing object destructing. But ig if we have to do this then it's aight too

jrosen081: I don't know of another way to have a string variable be the name of the argument. Is there a way to do it that I don't know?

cindy1u0: Oh wait I see what you are doing here cuz you have either id or imageURL. Nvm i think this is fine then lol. Potentially we can just create two success action since they are kinda doing different things

cindy1u0 (approved):

frontend/src/data/saga.js
jrosen081: So do you think this is a good abstraction, or should I go back to an unabstracted function? Which do you think is better?

cindy1u0: I would say to back to the unabstracted function :DD

jrosen081: Back to unabstracted!. LMK what you think

cindy1u0 (approved):
Gucci gang

PR #4, opened 2020-09-26
Prompt speech component

stefan-philip:


frontend/src/components/PromptSpeech.js
```diff
@@ -0,0 +1,50 @@
+import React from "react";
+import Speech from "react-speech";
+
+const PromptSpeech = (props) => {
```
cindy1u0: i think it would be better of doing ({ prompt }) here instead of (props) cuz later you can just do text={prompt} in your speech component

frontend/src/pages/Assessments.js
```diff
@@ -1,8 +1,15 @@
...
 import Layout from "../components/Layout";
+import PromptSpeech from "../components/PromptSpeech";
 
 const Assessments = () => {
-  return <Layout>Assessments</Layout>;
+  return (
+    <Layout>
+      <h1>Assessments</h1>
+      <p>Miniscule</p>
+      <PromptSpeech prompt={"miniscule"} />
```
cindy1u0: ya like here if you change the previous one to be ({ prompt }), it will just do prompt="miniscule." (u don't need to change anything here hehe)

frontend/src/components/PromptSpeech.js
cindy1u0: Woooo nice job creating a component for this xd

frontend/src/components/PromptSpeech.js
```diff
@@ -0,0 +1,50 @@
...
+    },
+    resume: {
+      hover: {},
+      button: {},
+    },
+  };
+
+  return (
+    <Speech
+      styles={speechStyle}
```
cindy1u0: Ah so actually we want to be consistent for the styles. It's prob better to do use styled component instead of having it to be styles={speechStyle}. A way to do it is to create a StyledSpeech before promptSpeech component. So something like:
```
const StyledSpeech = styled(Speech)`
  :hover {
    boxShadow: "0 0 15px 0 rgba(108,92,231,.69)",
  }
`;
```
Then instead of using <Speech> u can just use <StyledSpeech>. Then you don't need to set styles here

frontend/src/components/PromptSpeech.js
```diff
@@ -0,0 +1,50 @@
...
+      text={props.prompt}
+      pitch={1.05}
+      textAsButton={true}
+      displayText={"🔊"}
```
cindy1u0: Let's use SoundOutlined from antd design. So you can just write something like displayText={<SoundOutlined />} here instead of copying the speaker hehe

frontend/src/components/PromptSpeech.js
```diff
@@ -0,0 +1,50 @@
...
+    resume: {
+      hover: {},
+      button: {},
+    },
+  };
+
+  return (
+    <Speech
+      styles={speechStyle}
+      text={props.prompt}
```
cindy1u0: Yee so if you change the (props) to ({ prompt }), you can just do text={prompt} here.

frontend/src/pages/Assessments.js
```diff
@@ -1,8 +1,15 @@
 import React from "react";
 import Layout from "../components/Layout";
+import PromptSpeech from "../components/PromptSpeech";
 
 const Assessments = () => {
-  return <Layout>Assessments</Layout>;
+  return (
+    <Layout>
+      <h1>Assessments</h1>
```
cindy1u0, line 8: It doesn't matter for now since we need to change the text in the future too. But in the future, even the text doesn't have any styling, still make an attribute being like const Header = styled.h`` hehe. Same as the one below :0

frontend/src/components/PromptSpeech.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import Speech from "react-speech";
+import styled from "styled-components";
+import { SoundOutlined } from "@ant-design/icons";
+
+const StyledSpeech = styled(Speech)`
+  :hover {
+    boxshadow: "0 0 15px 0 rgba(108,92,231,.69)";
```
cindy1u0, line 8: We will prob need to change this part in the future

frontend/src/components/PromptSpeech.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+`;
+
+const PromptSpeech = ({ prompt }) => {
+  return (
+    <StyledSpeech
+      text={prompt}
+      pitch={1.05}
+      textAsButton={true}
+      displayText={<SoundOutlined />}
```
cindy1u0, line 18: We will prob need to change the text to the button stephanie designs in the future hehe

cindy1u0 (approved):

PR #5, opened 2020-09-27
Get all words from firebase

jrosen081:
image

### DIFF
--- a/frontend/src/data/actions.js
+++ b/frontend/src/data/actions.js
@@ -4,8 +4,8 @@ export const types = {
   ERROR: "ERROR",
   ADDUSER: "ADDUSER",
   DOWNLOADIMAGE: "DOWNLOADIMAGE",
-  GETWORDS: "GETWORDS",
-  GETWORDSSUCCESS: "GETWORDSSUCCESS",
+  GET_WORDS: "GETWORDS",
+  GET_WORDS_SUCCESS: "GETWORDSSUCCESS",
 };
 
 export const singleRequest = {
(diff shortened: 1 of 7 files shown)


frontend/src/data/actions.js
```diff
@@ -4,6 +4,8 @@ export const types = {
   ERROR: "ERROR",
   ADDUSER: "ADDUSER",
   DOWNLOADIMAGE: "DOWNLOADIMAGE",
```
cindy1u0, line 6: It's actually so minor tho, but is it okay if we change it to DOWNLOAD_IMAGE and then GET_WORDS/GET_WORDS_SUCCESS -> it's just easier to read in console for debugging

frontend/src/data/saga.js
```diff
@@ -52,3 +53,17 @@ function* watchDownloadImage(action) {
...
+function* watchGetWords(action) {
+  try {
+    let words;
+    const updateWithSuccess = async () => {
+      words = await firebaseInteractor.getWords();
+    };
+    yield call(updateWithSuccess);
+    yield put(singleRequest.getWordsSuccess({ words }));
+  } catch (error) {
+    console.log(error);
```
cindy1u0, line 66: i think it's aight not to put console.log(error) here since yield put sorta has it. But either way is fine

frontend/src/models/Word.js
```diff
@@ -0,0 +1,24 @@
+export default class Word {
```
cindy1u0, line 1: I wonder what if we do all the models in .ts heheh

cindy1u0 (approved):

frontend/src/data/saga.js
jrosen081: I had this when debugging, will remove lol

frontend/src/data/actions.js
jrosen081: I am going to be removing DOWNLOADIMAGE in a followup pr, so I will keep that for now, but will change GET_WORDS and GET_WORDS_SUCCESS

frontend/src/data/saga.js
jrosen081: Forgot to remove. I will make a big pr removing all console.logs prob next week in case something breaks and I need to try to debug firebase issues.

PR #8, opened 2020-10-03
Assessment skeleton

dillonhammer:
Assessments Landing Page

### DIFF
--- a/frontend/src/pages/Assessments.js
+++ b/frontend/src/pages/Assessments.js
@@ -1,23 +1,24 @@
-import React, { useState, useEffect } from "react";
+import React from "react";
 import styled from "styled-components";
 import Layout from "../components/Layout";
 import { Button } from "antd";
-import { INK } from "../constants/colors";
-
-const landingImageUrl =
-  "https://firebasestorage.googleapis.com/v0/b/vocab-buddy-53eca.appspot.com/o/assessments-landing.png?alt=media&token=168586ea-bb4e-499c-a711-5f3fe7a375b9";
+import { INK, INK_HOVER } from "../constants/colors";
+import { ASSESSMENTS_LANDING } from "../constants/images";
 
 const AssessmentsContainer = styled.div`
   display: flex;
-  font-size: 56px;
   font-weight: bold;
-  margin-top: 60px;
-`;
 
-const AssessmentsContainerMobile = styled.div`
-  font-size: 56px;
-  font-weight: bold;
-  width: 100%;
+  @media (max-width: 600px) {
+    flex-direction: column;
+    font-size: 56px;
+  }
+
+  @media (min-width: 601px) {
+    flex-direction: row;
+    font-size: 4vw;
+    margin-top: 60px;
+  }
 `;
 
 const Prompt = styled.div`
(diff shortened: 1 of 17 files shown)


frontend/src/pages/Assessments.js
```diff
@@ -1,13 +1,98 @@
...
+
+  useEffect(() => {
+    const resizeScreen = () => {
+      setScreenWidth(window.innerWidth);
+    };
+    window.addEventListener("resize", resizeScreen);
+    return () => {
+      window.removeEventListener("resize", resizeScreen);
+    };
+  }, []);
```
jrosen081: @cindy1u0 I wonder if we want to do screen size like this, or use CSS to do it with an `@media` tag. I know in my pr I used the `@media` tag which in my opinion looks cleaner, but in your pr you used `state` with width. What do you think?

frontend/src/pages/Assessments.js
```diff
@@ -1,13 +1,98 @@
...
     <Layout>
-      <h1>Assessments</h1>
-      <p>Miniscule</p>
-      <PromptSpeech prompt="miniscule" />
+      {screenWidth <= 600 ? (
+        <AssessmentsContainerMobile>
+          <LandingPageImage src={landingImageUrl} alt="assessments-landing" />
+          assessments
+          <Prompt>Select the correct images to match the target words.</Prompt>
+          <BeginButtonMobile block type="primary" size="large">
```
jrosen081: Are these css elements? If so, they should probably be in the `styled` components.

jrosen081 (approved):
This screen looks awesome. Just a few small things!

frontend/src/pages/Assessments.js
```diff
@@ -1,13 +1,98 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
+import styled from "styled-components";
 import Layout from "../components/Layout";
-import PromptSpeech from "../components/PromptSpeech";
+import { Button } from "antd";
+import { INK } from "../constants/colors";
+
+const landingImageUrl =
+  "https://firebasestorage.googleapis.com/v0/b/vocab-buddy-53eca.appspot.com/o/assessments-landing.png?alt=media&token=168586ea-bb4e-499c-a711-5f3fe7a375b9";
```
cindy1u0: I think i would put this in a constant file, and then import it at the top

frontend/src/pages/Assessments.js
cindy1u0: Oh ya I think for this, might be better to put it in the css component. In my PR, I had resize because there are some components we need to show or hide when the screen size is smaller than 600. I think you can make the container display: flex, then change the flex direction depending on how big the screens size is.
```
@media (max-width: 600px)`
   flex-direction: column;
`
```

frontend/src/pages/Assessments.js
cindy1u0: Not super sure tho but i think for type it has to be inline but for size u can prob put it in styled

dillonhammer:
@jrosen081 @cindy1u0 I refactored it a bit based on your feedback, lmk what you think

cindy1u0 (approved):
Oh ho ho looks gooood!

PR #9, opened 2020-10-08
Typescript migration

stefan-philip:


frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -40,22 +40,25 @@ export default class FirebaseInteractor {
...
-  async createAccount(email, password) {
+  async createAccount(email : string, password : string) {
     let userAuth = await this.auth.createUserWithEmailAndPassword(
       email,
       password
     );
+
+    // typescript is angry bc userAuth.user could be null
+    // @ts-ignore
     userAuth.user.sendEmailVerification();
```
jrosen081: Feel free to change this to:
```suggestion
    userAuth.user?.sendEmailVerification();
```
I think that would fix the issue.

frontend/src/models/types.ts
```diff
@@ -0,0 +1,48 @@
...
+  ERROR = "ERROR",
+  CREATE_USER = "CREATEUSER",
+  SIGN_IN = "SIGNIN",
+  AUTHENTICATION_SUCCESS = "AUTHENTICATIONSUCCESS",
+  GET_WORDS = "GETWORDS",
+  GET_WORDS_SUCCESS = "GETWORDSSUCCESS",
+};
+
+
+// i think we should make separate reducers to separate the different states if that makes sense
```
jrosen081, line 30: I definitely agree here. I don't think we got to a point of figuring out where each reducer lives, but I do fully agree we should split it up.

frontend/src/models/types.ts
```diff
@@ -0,0 +1,48 @@
...
+export interface LoginParams {
+  email : string;
+  password : string;
+}
+
+// i dont really like this
+// i think we should convert Word from a class to one of these types
+export interface WordList {
+  words: Word[];
+}
```
jrosen081: Feel free to do that in this pr imo. I think it makes sense

frontend/tsconfig.json
```diff
@@ -0,0 +1,30 @@
...
+    "noImplicitAny": true,
+    "strictNullChecks": true,
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "esnext"
+    ],
+    "allowJs": true, // should we make this false?
```
jrosen081, line 18: At some point probably yes I think, but is everything updated to ts? If so then I think we should now

jrosen081 (approved):
Wooooo! TYPES YAY!!!!

PR #10, opened 2020-10-10
Adding more to login, and also sign up

jrosen081:
This is not correctly styled, but has all of the moving parts lol. Will follow up with another pr later with styling updates.


frontend/src/components/Nav.tsx
```diff
@@ -96,7 +98,9 @@ const Nav : FunctionComponent = () : ReactElement => {
     return (
       <>
         <Link href="/assessments">assessments</Link>
-        <Link href="/interventions">interventions</Link>
+        <Link href="/interventions" onClick={() => firebase.auth().signOut()}>
```
cindy1u0, line 101: Can we create an action for this? Cuz i assume when you sign out you also want to change the reducer sign_in to be false right?

frontend/src/components/TextInput.tsx
```diff
@@ -0,0 +1,56 @@
...
+const TextInput = ({ value, onChange, text, type, expectedValue = null } : TextInputProps) : ReactElement => {
+  return (
+    <InputDiv>
+      <InputTitle>{text}</InputTitle>
+      <LoginInput
+        type={type}
+        value={value}
+        placeholder={text}
+        onChange={onChange}
+        style={
```
cindy1u0, line 22: You can actually add this part to LoginInput so that we don't need to have inline styling. Basically you have a prop being like expected={expected} and then in your Login input, it will look something like:

```
${(p: TextInputProps) => 
p.expectedValue && 'border: ... ... '}
```

frontend/src/data/saga.ts
```diff
@@ -1,16 +1,16 @@
...
-  singleRequest,
+  singleRequest
 } from "./actions";
-import {Action, ActionTypes, LoginParams, Word} from "../models/types";
 
 let firebaseInteractor = new FirebaseInteractor();
 
 export default function* rootSaga() {
-  yield all([root()]);
+  yield all([root(), waitForSignInAtStart()]);
```
cindy1u0, line 13: after deleting the single request in root, can you put waitForSignInAtStart() under root()? So like 

yield takeLatest(ActionTypes.WhateverRequest, waitForSignInAtStart);

frontend/src/data/saga.ts
```diff
@@ -1,16 +1,16 @@
...
 }
 
 function* root() {
```
cindy1u0, line 16: can we get rid of this it seemed like we don't rly need singlerequest anymore. Also can you delete the one in action too?

frontend/src/pages/CreateUser.tsx
```diff
@@ -0,0 +1,211 @@
...
+const CreateUser : FunctionComponent<CreateUserProps> = ({ signedIn, createUser }) : ReactElement => {
+  let [email, setEmail] = useState("");
+  let [password, setPassword] = useState("");
+  let [confirmPassword, setConfirmPassword] = useState("");
+  let [name, setName] = useState("");
+  let [accountType, setAccountType] = useState("");
+  let [age, setAge] = useState("");
+  let createUserWithCheck = () => {
+    if (confirmPassword !== password) {
+      alert("You need to confirm the password with the same password");
```
cindy1u0, line 115: Actually let's use antd's package Notification for alert :))

cindy1u0 (commented):
Man thanks for doing this chungus story

frontend/src/components/Nav.tsx
jrosen081: At some point yes, but I haven't actually thought enough about how sign out will actually work. I think that will end up being its own story, so I will punt on this for now if that is ok.

frontend/src/components/TextInput.tsx
jrosen081: I actually learned about this in my style pr. I might include that there since that does a lot of the correct styling. Does that work? If not, I definitely can add that here.

frontend/src/data/saga.ts
jrosen081: The only issue with that, is that then something needs to make the request (I think, not 100% sure if my mental model is correct). I did it in this style so that no matter where you start, you try to sign in. Does that make sense? I am not sure how to do it if I do it with the `takeLatest` style.

frontend/src/data/saga.ts
jrosen081: What do you want to get rid of? I am not sure which lines you are referring to.

frontend/src/pages/CreateUser.tsx
jrosen081: Oh didn't know it has it, might do that in the style pr as well, since that seems like a styling type of thing. Do you agree?

frontend/src/components/Nav.tsx
cindy1u0: okie

frontend/src/components/TextInput.tsx
cindy1u0: ohh ya sure add that in the other card then

frontend/src/data/saga.ts
cindy1u0: hmmm kinda? cuz the thing about root() is that since it contains all the takeLatest then we can just do everything all at once. Would takeEvery solve that problem?

frontend/src/data/saga.ts
cindy1u0: Ohhh tag the wrong line... line 26-31 in this file. The singleRequest.success() and singleRequest.error()

frontend/src/pages/CreateUser.tsx
cindy1u0: okie that makes sense :))

PR #13, opened 2020-10-20
Auto redirect users when logging in

jrosen081:
Few changes:
'/' now redirects to '/login' to make the naming simpler.

If you are unauthenticated, you get sent to '/login'. Once you sign in, you go back to the current page. In the browser history, it looks like you just went to the current page, without seeing that internal redirect.

All routes that need authentication should be an `AuthenticatedRoute` and not just a normal `Route`. This adds authentication checks on top of the normal route, and default redirect behavior.

Have a good day :)))))


stefan-philip (approved):

cindy1u0 (approved):
LOOKS GOOD CHUNGUS

PR #14, opened 2020-10-23
Assessment restructure

dillonhammer:
The Begin button Assessment page now starts the assessment. The url pattern was changed to be assessments/:id where the id now represents the assessment id, so it stays constant throughout the assessment. Right now there are only 3 words because I didn't feel like resizing and uploading more.

### DIFF
--- a/frontend/src/firebase/firebaseInteractor.ts
+++ b/frontend/src/firebase/firebaseInteractor.ts
@@ -103,9 +103,9 @@ export default class FirebaseInteractor {
   /**
    * Gets all possible words.
    *
-   * @returns { Promise<Array<Word>> }
+   * @returns { Promise<Assessment> }
    */
-  async getWords(): Promise<Assessment> {
+  async getAssessment(): Promise<Assessment> {
     let assessmentRef = await this.db.collection("assessments").get();
     let assessmentDocs = assessmentRef.docs;
     // Hardcoding first assessment for now
--- a/frontend/src/firebase/firebaseInteractor.ts
+++ b/frontend/src/firebase/firebaseInteractor.ts
@@ -107,7 +107,6 @@ export default class FirebaseInteractor {
    */
   async getWords(): Promise<Assessment> {
     let assessmentRef = await this.db.collection("assessments").get();
-    console.log("---------------------------------------------------");
     let assessmentDocs = assessmentRef.docs;
     // Hardcoding first assessment for now
     let assessment = assessmentDocs[0];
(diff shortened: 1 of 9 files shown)


frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -73,48 +79,57 @@ export default class FirebaseInteractor {
...
   }
 
   /**
    * Gets all possible words.
    *
    * @returns { Promise<Array<Word>> }
    */
-  async getWords() : Promise<Word[]> {
-    let wordRef = await this.db.collection("words").get();
+  async getWords(): Promise<Assessment> {
```
jrosen081: Can we name this something else? It seems weird to call it getWords and return an assessment

frontend/src/pages/Assessments/Assessments.tsx
```diff
@@ -56,11 +65,23 @@ const Assessments = () => {
...
           assessments
           <Prompt>Select the correct images to match the target words.</Prompt>
-          <PurpleButton text={"begin"} top={100} />
+          <PurpleButton
+            text={"begin"}
+            top={100}
+            onClick={() => {
+              // hardcoding assessment with id=1 for now
+              history.push("/assessments/1");
+              getWords();
```
jrosen081: Seems weird that we call this and don't actually use the assessment. I am not sure what the correct answer is lol, but this seems kind of weird to me.

frontend/src/pages/Assessments/Quiz.tsx
```diff
@@ -83,7 +82,7 @@ const ButtonContainer = styled.div`
 
 const connector = connect(
   (state) => ({
-    allWords: getAllWords(state),
+    assessment: getAllWords(state),
```
jrosen081: Same thing with `getAllWords` returning an assessment

jrosen081 (commented):
Looking really good. Just a question about the data flow and a few small naming nitpicks

frontend/src/pages/Assessments/data/reducer.tsx
```diff
@@ -1,25 +1,35 @@
...
+  ActionTypes,
+  Assessment,
+  AssessmentState,
+} from "../../../models/types";
+
+const emptyAssessment: Assessment = { id: -1, currentIndex: 0, words: [] };
 
 const initialState: AssessmentState = {
-  words: [],
+  assessment: emptyAssessment,
```
cindy1u0, line 12: I will think to initialize id, currentIndex, words in this initialState instead of just put assessment here. So like
```
   const initalState: AssessmentState = {
    id: 1,
    currentIdx: 0,
   words: [],
}
```
Also wondering why is ID -1 here?

PR #17, opened 2020-10-27
First intervention

cindy1u0:
First intervention. Follow up changes:
-> oral prompt after stefan pushes his changes
-> change the image url once the model the backend are done

### DIFF
--- a/frontend/src/components/PromptSpeech.tsx
+++ b/frontend/src/components/PromptSpeech.tsx
@@ -3,7 +3,6 @@ import React, {FunctionComponent, ReactElement} from "react";
 // @ts-ignore
 import Speech from "react-speech";
 import styled from "styled-components";
-import PlayButton from "../components/PlayButton";
 
 const SpeechContainer = styled.div`
   margin: 0;
@@ -15,16 +14,17 @@ const SpeechContainer = styled.div`
 
 interface SpeechProp {
   prompt: string;
+  button: ReactElement;
 }
 
-const PromptSpeech : FunctionComponent<SpeechProp> = ({ prompt }) : ReactElement => {
+const PromptSpeech : FunctionComponent<SpeechProp> = ({ prompt, button }) : ReactElement => {
   return (
     <SpeechContainer>
       <Speech
         text={prompt}
         pitch={1.05}
         textAsButton={true}
-        displayText={<PlayButton scale={0.8} />}
+        displayText={button}
       />
     </SpeechContainer>
   );
--- a/frontend/src/components/ReplayButton.tsx
+++ b/frontend/src/components/ReplayButton.tsx
@@ -1,0 +1,50 @@
+import React, { ReactElement } from "react";
+import styled from "styled-components";
+
+interface ReplayProps {
+  scale?: number,
+}
+
+const ReplayButtonContainer = styled.div`
+  display: flex;
+  padding: 5px;
+  width: max-content;
+`;
+
+const ReplayText = styled.p`
+  color: #000;
+  font-family: "Rubik";
+  font-weight: 700;
+  margin-bottom: 0;
+  margin-right: 5px;
+`;
+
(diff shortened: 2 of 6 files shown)


frontend/src/components/ReplayButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const ReplayButtonSVG = styled.svg`
+  ${(props) =>
+    `
+    transform: scale(${props.scale});
+  `}
+  path: hover {
+    cursor: pointer;
+  }
+`;
```
jrosen081: Do we want this as an svg, or just use an image and scale it?

frontend/src/components/ReplayButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+const ReplayText = styled.p`
+  color: #000;
+  font-family: "Rubik";
+  font-weight: 700;
+  margin-bottom: 0;
+  margin-right: 5px;
+`;
+
+const ReplayButtonSVG = styled.svg`
+  ${(props) =>
```
jrosen081: Can we destructure this?

frontend/src/pages/Interventions/FirstActivity.tsx
```diff
@@ -0,0 +1,101 @@
...
+  font-family: "Rubik";
+  font-weight: 700;
+  font-size: 18px;
+  margin-bottom: 10px;
+`;
+
+const WordTitle = styled.p`
+  font-family: "Rubik";
+  font-size: 35px;
+  font-weight: 700;
```
jrosen081, line 36: Should we make components that already have the fonts correct, so that we all use the same components, and then changing the font will change it for everyone?

frontend/src/components/PromptSpeech.tsx
```diff
@@ -15,16 +14,17 @@ const SpeechContainer = styled.div`
 
 interface SpeechProp {
   prompt: string;
+  button: ReactElement;
```
jrosen081, line 17: I wonder if we should make it like use the `children` input, so that like the PromptSpeech is like a wrapper around its component. What do you think?

jrosen081 (commented):
Looking awesome, just a few small comments!

frontend/src/components/ReplayButton.tsx
cindy1u0: I think this is gonna be an svg but then the svgs exported from figma are kinda big

frontend/src/pages/Interventions/FirstActivity.tsx
cindy1u0: As in like for the font size?

cindy1u0: I think it's okay to change the size instead of making the component. If it's the font-family, I think im just putting it to enforce it cuz i think in the index.js file rubik should be the one that's first found

frontend/src/components/PromptSpeech.tsx
cindy1u0: Hmm i think for promptSpeech it sounds more like component to me rather than a layout/wrapper cuz i dont think we need to do something like <PromptSpeech>blah blah </PrompSpeech> since technically only the prompt text is changing

frontend/src/components/ReplayButton.tsx
cindy1u0: oh yes yes

frontend/src/pages/Interventions/FirstActivity.tsx
jrosen081: ahhh makes sense nvm then

frontend/src/components/PromptSpeech.tsx
jrosen081: Ok sounds good!

jrosen081 (approved):

PR #23, opened 2020-11-07
Fourth intervention

dillonhammer:


jrosen081 (approved):
It looks like the FirstActivity and FourthActivity are _the same_, so do we think we can make a component that is the same, and then have FirstActivity and FourthActivity use that component?

cindy1u0 (approved):

PR #24, opened 2020-11-08
Select image

cindy1u0:
1. When an image is selected, the border will be highlighted -> intentionally did not included unselect cuz the user really has to pick an image anyways
2. Fixed image grid on different browser. Please QA on chrome, safari, and firefox (if u have)


frontend/src/components/ImagesComponent.tsx
```diff
@@ -0,0 +1,40 @@
...
+    cursor: pointer;
+    opacity: 0.8;
+  }
+
+  ${({highlight}: ImageProp) => highlight && `
+    border: 10px solid ${SEA_FOAM};
+  `};
+`;
+
+const ImagesComponent: FunctionComponent<ImagesProp> = ({ images })  : ReactElement => {
```
jrosen081: Can we rename this to something more specific to what it does? ImagesComponent doesn't really mean anything

frontend/src/components/ImagesComponent.tsx
```diff
@@ -0,0 +1,40 @@
...
+  let [selected, setSelected] = useState(-1);
```
jrosen081: I think this selected and setSelected should be one level above since there is no way that the parent component can see this state, which will be important later.

cindy1u0: ya so if the set select is in the parent component, then the shuffling behavior will be weird since every time u do onClick it rerenders. That's why I made this whole thing into a component. If we need to do any post request here, we can just use the idx here with passed in idx of the the current word index if necessary. I think the parent component doesn't need to see this state? I guess when will it need to?

frontend/src/components/ImagesComponent.tsx
cindy1u0: Okie! Any name suggestions?

stefan-philip (approved):

frontend/src/components/ImagesComponent.tsx
jrosen081: Sorry just getting around to this, umm "ImageGrid"?

frontend/src/components/ImagesComponent.tsx
jrosen081: It needs to know which was selected when sending this to firebase to know if it was right. Unless I am missing something.

cindy1u0: so ya i thought about it. To know which one is selected, it's just the selected variable right. To update the corresponding word, we can pass in the word ID a level up so that this ImagesComponent can have the word ID and make calls here

PR #26, opened 2020-11-14
Third activity

stefan-philip:


frontend/src/components/DelayedNextButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+  :hover {
+    background: ${INK_HOVER};
+  }
+
+  @media (max-width: 600px) {
+    width: 100%;
+  }
+`;
+
+const DelayedNextButton = ({ className = "", text, top, onClick, icon, delay}: NextButtonProps)  : ReactElement => {
```
cindy1u0, line 39: I think it's okay not to include className? Do we have to?

frontend/src/components/YesNoSelection.tsx
```diff
@@ -0,0 +1,92 @@
...
+  }
+  }
+}
+};
+  
+`;
+
+const YesNoSelection: FunctionComponent<YesNoSelectionProps> = ({correctAnswer}): ReactElement => {
+  let [selected, setSelected] = useState(-1);
+  const answer = correctAnswer ? 1 : 0;
```
cindy1u0, line 64: I think using True or false might be easier here?

frontend/src/components/YesNoSelection.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  return (
+      selected === answer ?
+          <YesNoContainer>
+            <YesNoButton highlight={1} onClick={() => setSelected(1)} text="yes" top={20}
+                         check={<CheckCircleFilled style={{color: SEA_FOAM}}/>}
```
cindy1u0: you can put the color in your styled-component :))

frontend/src/components/YesNoSelection.tsx
```diff
@@ -0,0 +1,92 @@
...
+};
+  
+`;
+
+const YesNoSelection: FunctionComponent<YesNoSelectionProps> = ({correctAnswer}): ReactElement => {
+  let [selected, setSelected] = useState(-1);
+  const answer = correctAnswer ? 1 : 0;
+
+  return (
+      selected === answer ?
```
cindy1u0, line 67: I am thinking, instead of using a ternary here, would it be a bit easier to just let the YesNoButton take in a t/f for highlight? So like, if selected === answer && selected -> Yes; if selected !== answer && selected -> No; if !selected -> nothing

cindy1u0 (changes requested):

frontend/src/components/YesNoSelection.tsx
stefan-philip: makes sense, will make changes soon!

frontend/src/components/DelayedNextButton.tsx
jrosen081: className is needed if you want to do `styled(DelayedNextButton)`. That is how the styling works. If not, then you do not need it.

frontend/src/components/YesNoSelection.tsx
stefan-philip: hm actually I'm not seeing a clear way to do this with boolean, since there are 3 cases

PR #27, opened 2020-11-15
Second intervention

dillonhammer:


frontend/src/pages/Interventions/SecondActivity.tsx
```diff
@@ -0,0 +1,220 @@
...
+              <Image
+                url={img.url}
+                correct={img.correct}
+                selected={index === selectedIndex}
+                onClick={() => setSelectedIndex(index)}
+              />
+              {!img.correct && index === selectedIndex && (
+                <RedCircle>
+                  <CloseOutlined
+                    style={{
```
cindy1u0: You can put the styling using styled-components. So like you can do something like 
```
const StyledCloseOutlined = styled(CloseOutlined)`
    font-size: blah blah
`
```

frontend/src/pages/Interventions/SecondActivity.tsx
```diff
@@ -0,0 +1,220 @@
...
+                      marginTop: "20px",
+                      color: "white",
+                    }}
+                  />
+                </RedCircle>
+              )}
+              {img.correct && index === selectedIndex && (
+                <GreenCircle>
+                  <CheckOutlined
+                    style={{
```
cindy1u0: same thing here

frontend/src/pages/Interventions/SecondActivity.tsx
```diff
@@ -0,0 +1,220 @@
...
+            {/* {title} */}
+            Miniscule
+          </WordTitle>
+          <Prompt>
+            <PromptSpeech
+              prompt="Miniscule. Touch the picture that shows miniscule."
+              button={<ReplayButton scale={0.8} />}
+            />
+          </Prompt>
+          {imageUrls.map((img, index) => (
```
cindy1u0, line 217: ahh i think we might need to shuffle it cuz like if not then the first one is always going to be true

cindy1u0 (changes requested):

cindy1u0 (approved):

PR #33, opened 2020-11-19
Connect interventions

cindy1u0:
Right now since second activity and third activity haven't been pushed yet, you can still test it on first and fourth activity


frontend/src/pages/Interventions/Activities.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const currentActivityIdx = interventions && interventions.activityIdx;
+  const wordList = interventions && interventions.wordList;
+  const activities = wordList && wordList[currentWordIdx].activities;
+  const title = wordList && wordList[currentWordIdx].word.value;
+
+  const ActivityComponent = (idx: number): ReactElement => {
+    switch (idx) {
+      case 0:
+        return (<FirstActivity title={title} imageUrl={activities.a1.url}/>);
+      case 1:
```
cindy1u0: will change here after second and third activities are pushed

frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -117,6 +117,22 @@ export default class FirebaseInteractor {
...
   }
 
+  /**
+   * update intervention idx number
+   */
+  async updateIntervention(
+    wordIdx: number,
+    activityIdx: number
+  ) {
+    this.unsubscribe?.apply(this);
```
jrosen081: This is only necessary for sign in, you do not need to unsubscribe here.

frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -117,6 +117,22 @@ export default class FirebaseInteractor {
...
+    let interventions = await this.db.collection("interventions").get();
+    let intervention = interventions.docs[0];
```
jrosen081: Should we have this function take in the id and update the intervention that corresponds to that id?

frontend/src/pages/Interventions/Activities.tsx
```diff
@@ -0,0 +1,59 @@
...
+const Activities: FunctionComponent<ActivityProps> = ({ interventions, getInterventions }): ReactElement => {
+  useEffect(() => {
+    if (!interventions) getInterventions();
+  }, [interventions, getInterventions]);
+
+  const currentWordIdx = interventions && interventions.wordIdx;
+  const currentActivityIdx = interventions && interventions.activityIdx;
+  const wordList = interventions && interventions.wordList;
+  const activities = wordList && wordList[currentWordIdx].activities;
+  const title = wordList && wordList[currentWordIdx].word.value;
```
jrosen081: I wonder if we should do something like we did on Assessment where there is one level that either shows a loading screen or the intervention, and that one gets the intervention. Also, should this get the intervention from the id?

frontend/src/pages/Interventions/FourthActivity.tsx
```diff
@@ -68,34 +77,48 @@ const ButtonContainer = styled.div`
...
 const FourthActivity = ({
   title,
   imageUrl,
+  interventions,
+  updateIntervention,
 }: FourthActivityProps): ReactElement => {
+  const activityIdx = interventions && interventions.activityIdx;
+  const wordIdx = interventions && interventions.wordIdx;
+  const nextActivityIdx = getNextActivityIdx(activityIdx);
+  // const nextWordIdx = getNextWordIdx(wordIdx);
```
jrosen081: This seems to be the same between Fourth and FirstActivity. I wonder if this should be passed in from the parent. Like I wonder if these should not worry about how to get data, but more just take the data and use it?

jrosen081 (changes requested):
Looking awesome! A few things about where data is stored/calculated

frontend/src/pages/Interventions/FourthActivity.tsx
cindy1u0: ohh got point i see what u mean!

frontend/src/pages/Interventions/Activities.tsx
cindy1u0: ohh for the loading screen i think i did that part later in the file

cindy1u0: Ah true but i see that right now intervention is fetched hard coded-ly lol i can prob change that

frontend/src/pages/Interventions/FourthActivity.tsx
cindy1u0: oh wait actually since the activity and word idx are changing, they need to be in the component instead of in the parent correct me if im wrong tho

frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -118,17 +118,39 @@ export default class FirebaseInteractor {
...
+    activityIdx: number
+  ) {
+    let interventions = await this.db
+      .collection("interventions")
+      .doc(setId)
+      .get();
+    let intervention = interventions.ref;
+    intervention.update({
+      wordIdx,
+      activityIdx,
```
jrosen081: You do not need to get here. Instead, you can just do `update` right away. Also, the update should be `await`ed so that it waits for the promise to resolve

frontend/src/pages/Interventions/Activities.tsx
```diff
@@ -0,0 +1,129 @@
...
+    if (!interventions) getInterventions();
+  }, [interventions, getInterventions]);
+
+  const setId = interventions && interventions.setId;
+  const currentWordIdx = interventions && interventions.wordIdx;
+  const currentActivityIdx =
+    interventions && interventions.activityIdx;
+  const wordList = interventions && interventions.wordList;
+  const activities = wordList && wordList[currentWordIdx].activities;
+  const title = wordList && wordList[currentWordIdx].word.value;
```
jrosen081: There should be a separate component that has the actual intervention that exists.

frontend/src/pages/Interventions/FirstActivity.tsx
```diff
@@ -1,16 +1,34 @@
...
+  activityIdx: number;
+  wordIdx: number;
+  prompt: string;
   imageUrl: string;
+  maxWordLength: number;
+  updateIntervention: ({
+    setId,
+    wordIdx,
+    activityIdx,
+  }: {
```
jrosen081: Maybe pass in a function that is literally `updateIntervention()`, and then the `Activities` DOM object will then do the updating of the index.

frontend/src/pages/Interventions/SecondActivity.tsx
```diff
@@ -156,61 +181,84 @@ const StyledCheckOutlined = styled(CheckOutlined)`
...
 
-const correctUrl = {
-  url:
-    "https://firebasestorage.googleapis.com/v0/b/vocab-buddy-53eca.appspot.com/o/jSyyDnxzx41VFQNQbbEw%2Fminiscule3.png?alt=media&token=cb4f4cf6-a1d0-465d-a972-087230d2ff05",
-  correct: true,
-};
+const CloudImageLeft = styled(CloudImage)`
+  position: absolute;
+  left: 0;
+  width: 15%;
```
jrosen081: I wonder if we want to make a like component that adds the cloud images so that all of these that need it gets the cloud from that Component, kind of like Layout now.

jrosen081 (changes requested):
Looking really good, just a few comments about some styling and division of responsibilities

jrosen081 (approved):
This looks perfect!

PR #34, opened 2020-11-21
Intervention Set created after Quiz

stefan-philip:


frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -114,31 +114,65 @@ export default class FirebaseInteractor {
...
+  async updateAssessment(id: string, responses: AssessmentResult[]) {
+    await Promise.all(responses.map(async (response) => {
+      return await this.db.collection("assessments").doc(id).collection("results").doc(response.word).set({correct: response.correct});
+    }))
+  }
+
+  async createInterventionSet(id: string, responses: AssessmentResult[]) {
+    await Promise.all(responses.map(async (response) => {
+
+      let word = (await this.db.collection("words").doc(response.word));
```
jrosen081, line 156: I wonder if we should make a helper function that given an id, gets the corresponding word?

frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -114,31 +114,65 @@ export default class FirebaseInteractor {
...
+      let incorrectURL = wordData.incorrectImages[Math.floor(Math.random() * wordData.incorrectImages.length)];
+      let incorrectURL2 = wordData.incorrectImages[Math.floor(Math.random() * wordData.incorrectImages.length)];
+
+      let activity3Answer = Math.random() > .5 ? true : false;
+      let activity3URL = activity3Answer ? correctURL : incorrectURL2;
+
+      let a1 = word.collection("intervention-set").doc("activity1").set({prompt: "Say " + wordValue, url: correctURL});
+      let a2 = word.collection("intervention-set").doc("activity2").set({correctUrl: correctURL, incorrectUrl: incorrectURL, prompt: "Click the image that is " + wordValue});
+      let a3 = word.collection("intervention-set").doc("activity3").set({correctAnswer: activity3Answer, prompt: "Is this image " + wordValue + "?", url: activity3URL});
+      let a4 = word.collection("intervention-set").doc("activity4").set({prompt: "Say " + wordValue, url: correctURL});
```
jrosen081, line 172: This is really cool. You are creating different sets per person. I love this. If you could comment on how each intervention is being created that would be awesome!

jrosen081, line 172: This seems to be creating the specific intervention set inside of the word object? I wonder if we want to actually create a new intervention (in the intervention collection), and assign its user id to the current user or something like that.

jrosen081 (changes requested):
This looks really cool. One or two comments about where to store the new intervention sets, and other than that looking amazing! Also, you probably want to rebase from master, hmu on slack if you want some help with that!

jrosen081:
Closing this as the spec was changed for the client.

PR #35, opened 2020-11-22
Store current index after each next click

jrosen081:


frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -153,10 +153,13 @@ export default class FirebaseInteractor {
     return { id, currentIndex, words: actualWords, firebaseId: assessment.id };
   }
 
-  async updateAssessment(id: string, responses: AssessmentResult[]) {
+  async updateAssessment(id: string, responses: AssessmentResult[], currentIdx: number) {
     await Promise.all(responses.map(async (response) => {
       return await this.db.collection("assessments").doc(id).collection("results").doc(response.word).set({correct: response.correct});
```
cindy1u0, line 158: i guess this is from before but is there a reason why this is set but not update?

frontend/src/pages/Assessments/Quiz.tsx
```diff
@@ -20,7 +20,7 @@ const connector = connect(
   }),
   {
     getAssessment: getAssessmentAction.request,
-    updateAssessment: (responses: AssessmentResult[], id: string) => updateAssessment.request({ responses, id })
+    updateAssessment: (responses: AssessmentResult[], id: string, isFinished: boolean, currentIdx: number ) => updateAssessment.request({ responses, id, isFinished, currentIdx })
```
cindy1u0: what i would do here since it's a connector would be just updateAssessment: updateAssessment.request,
then in line 44 i would do () => updateAssessment(.. .. .. )

cindy1u0 (changes requested):
Looks really nice :))) very clean heheh just a few comments

frontend/src/firebase/firebaseInteractor.ts
jrosen081: This one is set since there is a possibility that the actual doc does not exist. Update fails if the doc does not exist. Generally when creating something new you should use `set` and when updating use `update`

cindy1u0 (approved):
LGTM!

PR #37, opened 2020-11-22
Added blockers to all intervention activities.

jrosen081:


stefan-philip (approved):

frontend/src/pages/Interventions/SecondActivity.tsx
```diff
@@ -214,26 +215,30 @@ const SecondActivity = ({
...
-              )}
-              {img.correct && index === selectedIndex && (
-                <GreenCircle>
-                  <StyledCheckOutlined />
-                </GreenCircle>
-              )}
-            </ImageContainer>
-          ))}
+          <Blocker afterSeconds={15} message='Click on an image' repeatable={false}>
+            <div>
```
cindy1u0: let's just use <> and </> here since we don't rly need to use styled div

cindy1u0 (changes requested):
LGTM! Just one tiny tiny tiny change

frontend/src/pages/Interventions/SecondActivity.tsx
jrosen081: I tried this, but then the blocker does not get the actual click event. I think when we use <> </> then this is like a virtual element that ignores all events, so it does not get sent to the blocker. When there is this div, it then does get the event, which then makes it get the event. This behavior can be tested out here (https://vocab-buddy-1wvn7be9p.vercel.app/interventions/1), this uses not a div. Let me know if you know how I can change this to make it work.

cindy1u0 (approved):
xd

PR #38, opened 2020-11-30
Only highlight one button and change delay

jrosen081:


frontend/src/components/YesNoSelection.tsx
```diff
@@ -62,37 +69,20 @@ color: ${CORAL};
...
-              :
-              <>
-                <YesNoButton highlight={1} onClick={() => setSelected(1)} text="yes" top={20}
-                             check = {<CheckCorrect/>}
-                             icon={<LikeFilled/>}/>
-                <YesNoButton highlight={0} onClick={() => setSelected(0)} text="no" top={20}
-                             check = {<CheckIncorrect/>}
-                             icon={<DislikeFilled/>}/>
-              </>}
+      <YesNoButton highlight={selected === 0 ? correctAnswer ? 0 : 1 : -1} onClick={() => select(0)} text="no" top={20}
```
cindy1u0: so hmm the expected behavior would be: the kid should be able to select either Yes Or No, as in like they can still change their answer after picking one. If they answer it right, then you have the feedback icon (which is either a check mark or an X) once they select an answer. You can check the figma file. I think right now you are only allowed to select one and you can't change it

frontend/src/components/YesNoSelection.tsx
```diff
@@ -36,22 +33,32 @@ margin-right: 10px;
...
   `
   }
   }
-}
 };
 `;
 
+
+
 const CheckCorrect = styled(CheckCircleFilled)`
```
cindy1u0: For some reason this check circle filled is still white for me

cindy1u0 (changes requested):
Looks good but i think we should allow them to reselect their answer

cindy1u0 (approved):
NICESU

PR #42, opened 2021-01-31
redirected after signing up

cindy1u0:
Karmen and I added the getter for SignedIn in CreateUser.tsx


jrosen081 (approved):
Oops looked like I missed this one when signing up. Good catch!

PR #44, opened 2021-02-02
Update the blocker with updated design

jrosen081:
image


stefan-philip:
This looks great! One note: how do we feel about the rectangular selection box around the star when it is being dragged? It may be a good accessibility feature? idk really just thought I'd bring it up.

stefan-philip (approved):

jrosen081:
> This looks great! One note: how do we feel about the rectangular selection box around the star when it is being dragged? It may be a good accessibility feature? idk really just thought I'd bring it up.

Definitely a good point. I liked it for accessibility, but I will ask Stephanie her opinion!

frontend/src/components/Blocker.tsx
```diff
@@ -1,60 +1,158 @@
...
+import background from "../assets/slider.svg";
 
 interface BlockerProps {
-    children: ReactElement,
-    afterSeconds: number,
-    message: string,
-    repeatable: boolean
+  children: ReactElement;
+  afterSeconds: number;
+  repeatable: boolean;
```
sarah-witt, line 17: is there a reason why we need `repeatable`? it's always `false` for now.

sarah-witt (commented):
Looks good and works well for me!

frontend/src/components/Blocker.tsx
jrosen081: That is a good point. I believe at one point there was a discussion of whether they should be repeated, and don't remember the ending to that conversation. I am going to leave it now and we can ask the researchers if that is something they would want at the next client meeting.

cindy1u0 (approved):
LGTM!

frontend/package.json
```diff
@@ -11,10 +11,11 @@
     "@types/node": "^14.11.5",
     "@types/react": "^16.9.51",
     "@types/react-dom": "^16.9.8",
-    "antd": "^4.6.4",
+    "antd": "^4.7.0",
```
cindy1u0, line 14: it's prob fine but also we should make sure no other places break after we upgrade antd

sarah-witt (approved):

cindy1u0 (approved):

frontend/package.json
jrosen081: Based on my testing nowhere is broken

jrosen081:
@stefan-philip After  chatting with stephanie, I changed the color to sea foam and added a border radius!

PR #45, opened 2021-02-03
Assessment timer

sarah-witt:
Record the sum of the time spent on all words in an assessment


jrosen081 (approved):
This looks super awesome! One small suggestion would be to name it durationInSeconds so it is clear what units it is using.

angelinahan (approved):
this looks great! :)

stefan-philip (approved):

PR #46, opened 2021-02-08
Added a new hover color for the replay buttons

jrosen081:


frontend/src/constants/colors.ts
```diff
@@ -6,6 +6,7 @@ export const SKY = "#A2C5FB";
 export const CLOUD = "#E3EEFF";
 export const SEA_FOAM = "#68F2D2";
 export const LOGIN_BACKGROUND = "#E3EEFF";
+export const REPLAY_HOVER = "#B4D1FF";
```
cindy1u0: Looks good! I think this is fine. I was thinking like since we have the RGBA values, we can just do like a SKY(0.8) to change the opacity so that we don't need to add another color

cindy1u0 (approved):

frontend/src/constants/colors.ts
jrosen081: Are they technically the same color? It looks like it, but Stephanie's designs seem to imply they are 2 different colors.

stefan-philip: good point Cindy

cindy1u0: oh really? Time to @stephmiano -> should we use another color or just change the opacity of the current

stephmiano: reducing the opacity is fine if that is more straightforward, since yeah it's really just a lighter version of that color. it's the same situation as the purple primary buttons

PR #47, opened 2021-02-08
Disallow the next button when the intervention has not been interacted with

jrosen081:
This fixes the issue that @stefan-philip reported.

### DIFF
--- a/frontend/src/pages/Interventions/FirstActivity.tsx
+++ b/frontend/src/pages/Interventions/FirstActivity.tsx
@@ -100,6 +100,7 @@ const FirstActivity = ({
                 top={20}
                 delay={2000}
                 onClick={updateIntervention}
+                canBeShown
               />
             </Blocker>
           </ButtonContainer>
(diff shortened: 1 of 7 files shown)


frontend/src/pages/Interventions/FirstActivity.tsx
```diff
@@ -100,6 +100,7 @@ const FirstActivity = ({
                 top={20}
                 delay={2000}
                 onClick={updateIntervention}
+                canBeShown
```
cindy1u0, line 103: ig technically we don't really need canBeShown for first activity and fourth activity but I think it's okay to just leave it

cindy1u0 (approved):
LGTM ~

stefan-philip (changes requested):
On the yes-or-no activity, when you select yes or no the prompt is played again

jrosen081:
@stefan-philip Good catch. This should be fixed now

stefan-philip (approved):
Awesome!

PR #48, opened 2021-02-09
Create interventions from assessments.

jrosen081:
When the assessment is finished, this creates **8** new interventions, each with 3 words, assigns them to the user, and gives it a session.

### DIFF
--- a/frontend/src/firebase/firebaseInteractor.ts
+++ b/frontend/src/firebase/firebaseInteractor.ts
@@ -198,6 +198,30 @@ export default class FirebaseInteractor {
     });
   }
 
+  async createInterventionFromAssessment(
+    responses: AssessmentResult[]
+  ): Promise<void> {
+    let incorrectWords = responses.filter((response) => !response.correct);
+    incorrectWords.sort((a: any, b: any) => {
+      if (Math.random() > 0.5) return 1;
+      else return -1;
+    });
+    for (let i = 0; i < 8; i++) {
+      let wordList = [
+        incorrectWords[(i * 3) % incorrectWords.length].word,
+        incorrectWords[i * 3 + (1 % incorrectWords.length)].word,
+        incorrectWords[i * 3 + (2 % incorrectWords.length)].word,
+      ];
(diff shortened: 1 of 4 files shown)


stefan-philip (commented):
This looks great! Does isFinished currently get set to true when all of the assessment words are exhausted? How does it behave when the number of incorrect words is less than 24? 
- 6/7 incorrect words and then they finish the assessment
- 0 incorrect words and then they finish the assessment

jrosen081:
@stefan-philip Yes, `isFinished` is true in 2 cases:
1. They finish all words
2. They have too many wrong words (for right now it is 2 since we don't have 24 words lol, but will be updated to 24 later)

stefan-philip:
What would be created (if anything) when a user only gets 7 words wrong? (or 0)

jrosen081:
Right now the words wrap around (so it will show the same words over and over again). The researchers have said that the words are complicated enough that the kids should get 24 wrong certainly. I will ask that again at the next client meeting though to confirm.

stefan-philip (approved):

frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -198,6 +199,30 @@ export default class FirebaseInteractor {
...
+        incorrectWords[(i * 3 + 2) % incorrectWords.length].word,
+      ];
+      await this.db.collection("interventions").add({
+        activityIdx: 0,
+        wordIdx: 0,
+        wordList,
+        // Assign it to the current user
+        userId: this.auth.currentUser?.uid,
+        // Decide which session the intervention is in
+        interventionSession: i,
```
sarah-witt: nit: the naming `InterventionSession` is a bit confusing here since it makes me think of auth sessions. Would something like `interventionGrouping` make more sense?

jrosen081: I think just `session` would also work. Do you agree?

sarah-witt (approved):

PR #49, opened 2021-02-09
Added spacing / vertical height to login and signup butons

angelinahan:
Added padding to login and signup buttons to match the height of the form fields, and increased vertical margins around the buttons.

Screen Shot 2021-02-08 at 10 13 47 PM
Screen Shot 2021-02-08 at 10 13 39 PM


stefan-philip (approved):
Looks good!

PR #51, opened 2021-02-09
Create assessment after intervention completed

stefan-philip:
This PR includes redux actions that can get triggered when an intervention is finished. When the action is triggered, a new assessment is created with the words from the just-finished intervention, and the intervention stored in the state is cleared out.

NEW: The action is triggered when the intervention is finished


frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -133,6 +133,43 @@ export default class FirebaseInteractor {
...
+      let wordData = await this.getWord(word);
+
+      let newWordData = {
+        correctImage: wordData.correctImage,
+        dateCreated: wordData.createdAt,
+        incorrectImages: wordData.incorrectImages,
+        value: wordData.value,
+      };
+
+      await newAssessment.collection("words").doc(word).set(newWordData);
```
jrosen081: I don't think you need to override the word. The word already has these values, so I think you are just overriding this with the same data.

frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -133,6 +133,43 @@ export default class FirebaseInteractor {
...
+    let intervention = await this.db
+      .collection("interventions")
+      .doc(setId)
+      .get();
+    let interventionData = intervention.data();
+    if (interventionData == null) {
+      throw new Error(`Intervention with id ${setId} does not exist`);
+    }
+
+    let wordList = interventionData.wordList;
```
jrosen081: While this looks super great, I think that you are supposed to have the new assessment's words be the ones that were incorrect from the previous **assessment**, not intervention. Since we technically don't have a previous assessment right now, you can probably get one hardcoded assessment and do it from there.

jrosen081 (changes requested):
The redux actions/firebase looks great, but we want to get the words from a previous assessment.

frontend/src/firebase/firebaseInteractor.ts
stefan-philip: Oh right, awkward 😅

frontend/src/firebase/firebaseInteractor.ts
stefan-philip: Would this mean that after EACH of the 8 interventions created for an assessment, there is the same assessment (with only incorrect words) recreated?

jrosen081: Yes (like the assessment gets smaller as they get words correct)

sarah-witt: Updated to hardcode a previous assessment and add the incorrect words to the new intervention!

frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -134,6 +134,39 @@ export default class FirebaseInteractor {
...
+      .doc(assessmentId)
+      .collection("results")
+      .get();
+    if (results == null) {
+      throw new Error(`Assessment with id ${assessmentId} does not exist`);
+    }
+
+    let wordList: string[] = [];
+    results.forEach(function (result) {
+      if (!result.get("correct")) {
```
jrosen081: It generally makes more sense to call `data` on each result, which gives a plain javascript object, that you can call `.correct` on. This works as  well tho

jrosen081 (approved):
Looks amazing, just a small comment and then :shipit:

frontend/src/firebase/firebaseInteractor.ts
sarah-witt: sure, updated!

PR #52, opened 2021-02-09
Assessment Reward Page

karmenlu:
[As a kid, I want to see an assessment reward page.](https://www.notion.so/As-a-kid-I-want-to-see-an-assessment-reward-page-c1ba433f4b254b199aa3ba18f5c9d58b)
I created a basic layout of the assessment reward page. The buttons directs to the dashboard. Things should be resized/recolored.

To-do:

- [x] CloudGroup does not span entire height on mobile. Clouds scrunched at the top.
- [x] PurpleButton does not span entire width on desktop view.
- [x] PurpleButton should be recolored.
- [x] direct to reward page after two questions wrong.

### DIFF
--- a/frontend/src/components/PurpleButton.tsx
+++ b/frontend/src/components/PurpleButton.tsx
@@ -10,6 +10,7 @@ interface PurpleButtonProps {
   check?: ReactElement;
   icon?: ReactElement;
   className: string;
+  padding?: number;
 }
 
 const ButtonContainer = styled(Button)`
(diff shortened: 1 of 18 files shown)


frontend/src/components/PurpleButton.tsx
```diff
@@ -47,12 +49,14 @@ const PurpleButton = ({
   onClick,
   icon,
   check,
+  padding,
```
jrosen081, line 52: I think if you use a `styled-component` you can add padding there!

jrosen081 (approved):
Looks great other than one small comment about `styled-components`

frontend/src/components/PurpleButton.tsx
cindy1u0: ya actually that's true. Also we were thinking that maybe in the future if we have other buttons with longer words, then we might need this padding to adjust the purple area

PR #55, opened 2021-02-16
Fix isssue with images disappearing

jrosen081:
There was a bug where the images would go away and disappear and reappear. This fixes that bug (and also fixes a resizing issue that looked weird). Go to second activity and click around with the example/non-example screen. Make sure there are no regressions.


cindy1u0 (approved):
LGTM :))

PR #56, opened 2021-02-16
Load user's current intervention/assessment

angelinahan:
This PR adds a few changes related to making sure the user is able to see only their current assessment/intervention

- When a user signs up, an assessment is created with all the words and assigned to them
- We now keep track of a user's current intervention / assessment and load the one they are on
- If a user tries to start to an assessment when they are on an intervention & vice versa, we display an error banner

Screen Shot 2021-02-16 at 6 34 36 PM


frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -115,10 +124,35 @@ export default class FirebaseInteractor {
...
+  async updateCurrentUser(user: Partial<User>) {
+    this.db.collection("users").doc(this.auth.currentUser?.uid).update(user);
+  }
+
+  async createInitialAssessment() {
+    let words = await this.db.collection("words").get();
+    let wordIds = words.docs.map((word) => word.id);
+    let newAssessmentFields = {
+      currentIndex: 0,
+      // do we need to include in the beginning?
```
sarah-witt: It's not needed but also wouldn't hurt to have for clarity https://firebase.google.com/docs/reference/js/firebase.firestore.FieldValue#static-increment

frontend/src/pages/Assessments/data/reducer.ts
```diff
@@ -24,16 +26,30 @@ const assessmentReducer = (
...
           words: payload.assessment.words,
           firebaseId: payload.assessment.firebaseId,
+          sessionId: payload.assessment.sessionId,
         },
         finished: false,
       };
     case ActionTypes.UPDATE_ASSESSMENT_SUCCESS:
       let { isFinished }: UpdateAssessmentSuccess = payload;
       return {
-        ...state,
```
sarah-witt, line 33: why was this removed?

jrosen081: If you are updating the assessment, you should not need the error or the current assessment id again, so we removed it.

frontend/src/firebase/firebaseInteractor.ts
jrosen081: Where would this be used?

sarah-witt: Sorry if I wasn't clear- I was responding to the comment in the code. Since we use increment for `durationInSeconds` and increment sets a value if the field does not exist, then a default value of 0 isn't needed when the assessment is created,  but specifying it at assessment creation time wouldn't hurt because it doesn't affect the value and makes it more clear in the code that `durationInSeconds` is an assessment field if that makes sense.

jrosen081: Ah gotcha makes sense. I will keep it for clarity

sarah-witt: Sounds good! Also can you remove or update the comment?

jrosen081: I will remove it

sarah-witt (approved):
looks good to me!

PR #60, opened 2021-02-28
Set up interventions to have three yes/no questions

jrosen081:


frontend/src/pages/Interventions/ActivitiesComponent.tsx
```diff
@@ -61,9 +61,32 @@ const ActivityComponent: FunctionComponent<ActivityComponentProp> = ({
           imageUrl={activities.a3.url}
           answer={activities.a3.correctAnswer}
           updateIntervention={updateIntervention}
+          key={0}
```
jrosen081, line 64: This key forces react to rerender, which resets the state of the `ThirdActivity` component. If this is not there, the state does not reset, and it looks really weird.

cindy1u0: this is interesting

cindy1u0 (approved):
LGTM! It works well on preview

PR #61, opened 2021-03-01
Add intervention duration

sarah-witt:


frontend/src/pages/Interventions/data/actions.ts
```diff
@@ -44,6 +44,7 @@ interface UpdateInterventionAction {
   wordIdx?: number;
   activityIdx?: number;
   interventions?: Interventions;
+  durationInSeconds?: number;
```
jrosen081: Is there ever a time where this will be undefined? I wonder if we want this to be optional or not

jrosen081 (approved):
Looks great and works super well for me!

frontend/src/pages/Interventions/data/actions.ts
sarah-witt: We were actually wondering this too- we noticed all of the fields of all of the intervention actions were optional so we just made `durationInSeconds` optional to follow the pattern. If we make it not optional then should we also change the other fields?

cindy1u0 (approved):

frontend/src/pages/Interventions/data/actions.ts
jrosen081: Yes (but keep answer2Correct, and answer3Correct optional since those fields are actually optional)

stefan-philip (approved):

jrosen081 (approved):

cindy1u0 (approved):
LGTM

PR #62, opened 2021-03-02
changed to no closecircle

cindy1u0:
Switched up the logic since we want checkcircle for no if the correctAnswer is false, meaning the pic doesn't represent the word
Screen Shot 2021-03-01 at 9 29 19 PM
Screen Shot 2021-03-01 at 9 29 16 PM
Screen Shot 2021-03-01 at 9 27 39 PM
Screen Shot 2021-03-01 at 9 27 34 PM

### DIFF
--- a/frontend/src/components/YesNoSelection.tsx
+++ b/frontend/src/components/YesNoSelection.tsx
@@ -87,10 +87,10 @@ const YesNoSelection: FunctionComponent<YesNoSelectionProps> = ({
           top={20}
           check={
             selected === 0 ? (
-              !correctAnswer ? (
-                <CheckCircleFilled style={{ color: SEA_FOAM }} />
-              ) : (
+              correctAnswer ? (
                 <CloseCircleFilled style={{ color: CORAL }} />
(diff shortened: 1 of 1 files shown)


frontend/src/components/YesNoSelection.tsx
```diff
@@ -86,10 +87,10 @@ const YesNoSelection: FunctionComponent<YesNoSelectionProps> = ({
           top={20}
           check={
             selected === 0 ? (
-              correctAnswer ? (
-                <CheckCircleFilled style={{ color: CORAL }} />
-              ) : (
+              !correctAnswer ? (
```
jrosen081: Any chance we can flip this around so it is like `correctAnswer ? ... : ...`? I think that is more readable since the true case is first.

jrosen081 (approved):
:ship: :rocket: 💎

frontend/src/components/YesNoSelection.tsx
cindy1u0: okie

PR #64, opened 2021-03-07
disabled resubmit for second and third activity

cindy1u0:


frontend/src/pages/Interventions/SecondActivity.tsx
```diff
@@ -166,6 +166,7 @@ const SecondActivity = ({
   updateIntervention,
 }: SecondActivityProps): ReactElement => {
   const [selectedIndex, setSelectedIndex] = useState(-1);
+  const [hasSelected, setHasSelected] = useState(false);
```
jrosen081: I don't think we need this, you can just check if the selectedIndex === -1 or not and that is the same thing, with less state to manage.

jrosen081 (approved):
🐐

PR #65, opened 2021-03-08
Wrote script to upload words to firebase

jrosen081:
There is a lot technically in this pr, but the only important part is `script.js`. The rest are just examples so you can test the pr. See [here](https://www.notion.so/Upload-Strategy-For-Words-48b350143e404a7a982b44ad5161b308) for the naming strategy.


sarah-witt (commented):
This looks so great! One thing I noticed is that some activities have multiple prompts that are to be said at different times, like one is said at the beginning of the activity, and one is said after the user has pressed a button. We would probably need to store the prompts separately in this case (like in a list or something). Another thing I am not sure about is if the assessment prompts will ever vary across words. If this is so, then we would also need to store assessment prompts in firebase instead of hardcoding the prompt in the front end like we do now.

jrosen081:
> This looks so great! One thing I noticed is that some activities have multiple prompts that are to be said at different times, like one is said at the beginning of the activity, and one is said after the user has pressed a button. We would probably need to store the prompts separately in this case (like in a list or something). Another thing I am not sure about is if the assessment prompts will ever vary across words. If this is so, then we would also need to store assessment prompts in firebase instead of hardcoding the prompt in the front end like we do now.

For your first comment, where did you see that? I was under the impression there was one prompt that we are just repeating. For the second comment, I am under the impression all of the assessment prompts all follow the same pattern, with a different word. If that does change (from feedback from the clients when I send the actual strategy), we can update the code! What do you think?

sarah-witt:
> For your first comment, where did you see that? I was under the impression there was one prompt that we are just repeating. 

I saw it in the example slides the researchers sent to us. For example, in "construct," the third activities all have multiple prompts.

>For the second comment, I am under the impression all of the assessment prompts all follow the same pattern, with a different word. If that does change (from feedback from the clients when I send the actual strategy), we can update the code! What do you think?

Sounds good!

jrosen081:
@sarah-witt Good catch! Looks like we probably need to rework a bit about how the prompts work at the moment. I think what  I have here will work with all of the "beginning" prompts, and then we can add another small script to backfill the extra info when we rewrite some of the logic relating to the "ending" prompts. Beginning being what happens at the beginning of the activity and ending what happens at the end. What do you think?

sarah-witt (approved):
Sounds good. The script looks good to me!

stefan-philip:
With the prompt text files, Sarah and I were originally thinking that the researchers would write the prompts on a Google Doc file. When the files are downloaded, these are downloaded as .docx files.

Would it be a hassle for the researchers to work on .txt files or is that just something that we'll require?

jrosen081:
@stefan-philip We can download it as a txt from google drive so if they end up giving us google drive docs, we can format it correctly. I know on Mac you can use TextEdit which will give out txt files. I did update the doc to specify txt, but If they do end up giving us google docs, we can do the transition manually. Does that sound reasonable?

stefan-philip:
> @stefan-philip We can download it as a txt from google drive so if they end up giving us google drive docs, we can format it correctly. I know on Mac you can use TextEdit which will give out txt files. I did update the doc to specify txt, but If they do end up giving us google docs, we can do the transition manually. Does that sound reasonable?

Sounds good! Thanks for this PR 🦜

stefan-philip (approved):

PR #66, opened 2021-03-11
Weekly progress

angelinahan:
- added field to firebase to keep track of user's daily activity
- display weekly progress tracker on dashboard

We had a question for @stephmiano and anyone else who has thoughts on this - on smaller mobile screens the dates start spilling out of the container because there isn't enough room. (see the last screenshot where the screen is 425px). Should we wrap the dates in this case? If yes, would the preference be to wrap it evenly so there's 4 on one row and 3 on the other, or keep the first row as full as possible and only wrap as necessary (e.g. 6 on the first row 1 on the second), or another arrangement?

Screen Shot 2021-03-10 at 8 39 33 PM

Screen Shot 2021-03-10 at 8 55 47 PM

Screen Shot 2021-03-10 at 8 50 45 PM


frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -100,7 +100,9 @@ export default class FirebaseInteractor {
     userAuth.user.sendEmailVerification();
     if (accountType === "STUDENT") {
       let initialAssessmentId = await this.createInitialAssessment();
+      let currentDaysActive = new Array<String>();
```
jrosen081: Can we use the array literal syntax here?
```suggestion
      let currentDaysActive = [];
```

frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -214,6 +218,42 @@ export default class FirebaseInteractor {
...
+      if (today !== day && (dayOfWeek > todayDayOfWeek || daysApart >= 7)) {
+        clearPreviousWeek = true;
+      }
+    });
+
+    if (clearPreviousWeek) {
+      await this.db.collection("users").doc(userId).update({
+        daysActive: [],
+      });
+    }
```
jrosen081: Do we inherently need to clear the previous week? I think it would be better to keep the values so that we can use the previous week/month's usage for stats/insights about app usage. If we did not clear it, would that break the current implementation?

frontend/src/models/types.ts
```diff
@@ -93,6 +93,7 @@ export interface User {
   sessionId: SessionId;
   onAssessment: boolean;
   currentInterventionOrAssessment: string;
+  daysActive: Array<string>;
```
jrosen081: Can we use the literal syntax here? 
```suggestion
  daysActive: string[];
```

frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -238,19 +240,57 @@ const ProgressStatDescription = styled.p`
   }
 `;
 
+const DayLabel = styled.p<DayLabelType>`
+  font-weight: ${(props) => (props.isToday ? 700 : 400)};
```
jrosen081: Can we destructure this please? Also did not know you could do styledcomponents with the type there, I think that looks super nice!!

jrosen081 (changes requested):
This looks good, but I think we should store the full date, so that we can use the data later. In regards to your question, we should probably scale it down, so that it continually gets smaller (including the text) as we get less space.

frontend/src/pages/Dashboard/Dashboard.tsx
stefan-philip: An example of destructuring in a styled component can be found in the YesNoSelection component

frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -354,9 +441,16 @@ const Dashboard: FunctionComponent<DashboardParams> = ({
...
-
             <WeekContainer>
-              <p>stars!</p>
+              {dayLabels.map((label: string, index: number) => {
+                return (
+                  <DayOfWeek
+                    name={label}
+                    day={index}
+                    currentUser={currentUser}
+                  ></DayOfWeek>
```
stefan-philip: Is the closing tag necessary?

frontend/src/models/types.ts
stefan-philip: string[] if that wasn't clear 😂

stephmiano:
Yes if I'm not mistaken I believe the calendar can be sized down for mobile so that it fits in one row, including the container itself, text, stars, and spacing. The mobile dashboard wireframe can be referenced for relative sizing of the calendar on mobile.

frontend/src/firebase/firebaseInteractor.ts
sarah-witt: We were thinking that storing all the dates for every user would be a lot of data to store. Would you suggest clearing after a month? Right now if we did not clear it it would break the current implementation, but it's something we could do.

sarah-witt:
> This looks good, but I think we should store the full date, so that we can use the data later. 

That was our original plan, but firebase stores javascript dates as firebase Timestamps, which only have seconds and nanoseconds (and no time zone). So even if we store dates we will still have to convert the data to javascript dates (and somehow remove the time portion of the date).

frontend/src/firebase/firebaseInteractor.ts
jrosen081: I honestly don't think we need to clear it at all. It probably just going to be a small string, so probably nothing to worry about

jrosen081:
@sarah-witt You could store the string representation, such as "2018-08-20" (or some other representation) which would have the day for you? Would that work do you think?

sarah-witt:
> @sarah-witt You could store the string representation, such as "2018-08-20" (or some other representation) which would have the day for you? Would that work do you think?

We're currently storing it as a string date from the toString on javascript's Date(), ie "Mon Mar 08 2021 00:00:00 GMT-0500 (Eastern Standard Time)" since it has the timezone on it. Is that fine?

jrosen081:
Sure!

jrosen081 (approved):
Looks amazing!!

PR #68, opened 2021-03-14
Added error messages/screens throughout the app

jrosen081:
This adds error toasts to all landing pages and sign up/in pages when an error occurs and adds an error screen for a 404 and any loading errors on Assessments/Interventions/Dashboard.

image


angelinahan (changes requested):
Looks nice!! I think you're missing `error: getCreatUserError(state)` on line 145 of CreateUser.tsx - right now the network errors for creating a user aren't showing up because the connector isn't getting them. Also the "create" is missing an e in the function name, super minor thing but if you're going back and changing things anyway, might as well! other than that looks good!

jrosen081:
@angelinahan Good call on both of those! Updated!

angelinahan (approved):
yay, looks good!!

PR #69, opened 2021-03-14
Make sure the person goes through only when getting 24 wrong

jrosen081:
We actually want the kids to get 24 wrong, not 2, so updating this.


cindy1u0 (approved):
thank u soldier

PR #70, opened 2021-03-15
Add stats to the dashboard

jrosen081:


frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -487,14 +495,25 @@ const Dashboard: FunctionComponent<DashboardParams> = ({
...
+              />
+              <Stat
+                number={currentUser.sessionId + 1}
                 description={"assessments completed"}
               />
               <Stat
-                number={3}
+                number={
+                  currentUser.sessionId +
+                  (currentUser.onAssessment ? 1 : 0)
```
stefan-philip, line 528: could you explain this? I'm not quite getting it, maybe I'm confused about what sessionId is.

jrosen081: So sessionId is the current session. -1 stands for pre-assessment, and then goes through 9. You have always done sessionId + 1 assessments since you had to do 1 assessment per session before the current one. You have done sessionId interventions, since we don't have one for -1. If you are on an assessment, then you did the intervention for the current session, so you have done sessionId + 1 interventions. Does that make sense?

stefan-philip: Got it, makes sense!

stefan-philip (approved):

PR #76, opened 2021-03-21
Fix sizing of dashboard so it does not overflow on laptop screens

jrosen081:
image


cindy1u0 (approved):
ty soldier

frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -169,11 +167,12 @@ const SessionCardContainer = styled.div`
   display: grid;
   grid-template-columns: 1fr 1fr 1fr 1fr;
   min-width: 100%;
-  grid-gap: 32px;
+  grid-column-gap: 3vh;
```
stefan-philip: Maybe we should use vmin? The row and column gaps are uneven right now

stefan-philip: Why not use `gap` ?

frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -152,10 +153,7 @@ const SessionContainer = styled.div`
   align-items: flex-start;
 
   flex: 3;
-  min-height: 100%;
-  padding-left: 64px;
-  padding-top: 64px;
```
stefan-philip, line 159: We probably want to keep some top padding based on the design

frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -229,11 +228,9 @@ const NextSessionButton = styled(PurpleButton)`
 
 const WeekProgressContainer = styled.div`
   flex: 2;
-  min-height: 100%;
 
-  padding-left: 64px;
-  padding-right: 64px;
-  padding-top: 64px;
```
stefan-philip, line 238: Probably want some top padding from design

frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -255,14 +252,14 @@ const ProgressBox = styled.div`
...
-  padding: 30px 30px 30px 30px;
+  padding: 2vh 2vw;
 
   flex: 1 0 50%;
   order: 0;
   flex-grow: 1;
 
   max-width: calc(50% - 16px);
-  margin-bottom: 32px;
+  margin-bottom: 2vh;
```
stefan-philip: Hmmm this could definitely be saved for a future design bug ticket, but I think it would look cleaner if these progress boxes used margin-top, so that we can align the whole ProgressBoxContainer with the bottom of the screen, instead of a small gap. It looks inconsistent with the bottom of the SessionCards

frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -182,7 +181,7 @@ interface SessionCompletionProp {
 }
 
 const SessionBox = styled.div`
-  padding: 40px;
+  padding: 1.5vh 1.5vw;
```
stefan-philip: My thoughts on the session box:
- may need more padding
- maybe make the image smaller
- make the font size smaller
- probably want to do justify-content: flex-end; I'm getting a weird thing where "session 2" text is positioned higher than the rest of the texts 

Not necessarily something you need to fix in this PR, but just noting it for a bug ticket or something

stefan-philip (commented):
Jack you're a king. Made a few comments about padding and grid gaps, and some notes for design related bugs. But this is really awesome, thank you!!

frontend/src/pages/Dashboard/Dashboard.tsx
jrosen081: They were different sizes, but using vmin I can use gap!

frontend/src/pages/Dashboard/Dashboard.tsx
jrosen081: Good catch! I am working to fix these (and I am pretty sure I introduced this bug, so I am fixing it :)

jrosen081:
@stefan-philip Good catches on everything you commented on. Can you take another look and make sure it looks better?

frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -203,14 +205,15 @@ const SessionNumber = styled.p`
...
+  margin-bottom: 0;
 
   @media (max-width: 900px) {
     font-size: 3vw;
   }
 `;
 
 const SessionImage = styled.img`
-  max-width: 100%;
+  max-width: 60%;
```
stefan-philip, line 216: Something odd is going on here. If you resize your screen to just above 900 px, the session 1 image is much larger than the rest. It makes the vertical spacing odd for the top row. Could be a bug ticket though

Screen Shot 2021-03-21 at 10 13 42 PM

stefan-philip (approved):
Great work!

frontend/src/pages/Dashboard/Dashboard.tsx
stefan-philip: oops my bad! looks great now

frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -231,11 +233,10 @@ const NextSessionButton = styled(PurpleButton)`
 
 const WeekProgressContainer = styled.div`
   flex: 2;
-  min-height: 100%;
```
stefan-philip, line 236: Can we use flex here so that we can do `justify-content: flex-end` ?  "this week + container" and the "your progress + container" are the children elements (they are 4 separate elements as of right now, but can wrap them in divs to group two)

I think it would look better if the "this week + container" was aligned with the bottom of the screen

stefan-philip: Sorry disregard this. I don't know if this comment accurately represents Stephanie's design

frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -327,7 +319,19 @@ const DayLabel = styled.p<DayLabelType>`
 const Star = styled.img`
   height: 51px;
```
stefan-philip, line 320: so sorry for being nit picky.. The star sizing at like 900px screen size goes to the edge of the container. Maybe instead of px use vw/vh ?

Screen Shot 2021-03-21 at 10 37 53 PM

frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -126,6 +128,7 @@ const TitleText = styled.p`
   font-weight: 700;
```
stefan-philip, line 128: We may want to make the font size for the titles  in terms of vh/vw

jrosen081: I think the titles are good to stay at the full size, since they are like an important part if that makes sense.

stefan-philip: That's good with me!

frontend/src/pages/Dashboard/Dashboard.tsx
jrosen081: Hmmm this is interesting. I have **no** idea why it is happening, so I won't fix it in this pr. Do you want to make a bug ticket for it on Notion?

frontend/src/pages/Dashboard/Dashboard.tsx
jrosen081: This has been updated!

frontend/src/pages/Dashboard/Dashboard.tsx
stefan-philip: No prob! Will put up a bug card later tonight

PR #77, opened 2021-03-21
Get rid of manual moving around the app (no more assessments/interventions in the nav bar)

jrosen081:
image


stefan-philip (approved):
Looks great!

PR #79, opened 2021-03-21
User icon

cindy1u0:
This PR allows a user to update its profile picture
https://user-images.githubusercontent.com/45154128/111921277-27925500-8a6a-11eb-9e53-9c161cf5668c.mov


stefan-philip (approved):
AWESOME!

frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -133,12 +151,26 @@ export default class FirebaseInteractor {
...
       throw new Error("Invalid user");
     }
   }
 
+  /**
+   * Update user's profile icon; creates the field for profile icon if not already exists
+   * @param {String} url: the profile icon url
+   */
+  async updateUserProfileIcon(url: string) {
+    await this.db.collection("users").doc(this.auth.currentUser?.uid).set(
```
jrosen081: I think we should use `update` here so we don't need the second parameter. If we use `update` it has the same behavior, but is a bit clearer in my opinion.

frontend/src/pages/Dashboard/data/actions.ts
```diff
@@ -36,3 +36,18 @@ export const GetData = {
...
+  request: (url: string) => ({
+    type: ActionTypes.CHANGE_PROFILE_ICON_REQUEST,
+    payload: { url },
+  }),
+  success: (url: string) => ({
+    type: ActionTypes.CHANGE_PROFILE_ICON_SUCCESS,
+    payload: { url },
+  }),
+  error: (error: Error) => ({
+    type: ActionTypes.CREATE_USER_ERROR,
```
sarah-witt: should this be `CHANGE_PROFILE_ICON_ERROR`?

frontend/src/pages/Dashboard/data/reducer.ts
```diff
@@ -52,6 +52,16 @@ const dashboardReducer = (
         ...state,
         error: action.payload.error,
       };
+    case ActionTypes.CHANGE_PROFILE_ICON_SUCCESS:
+      return {
+        ...state,
+        user: state.user
```
jrosen081, line 58: I don't think you can get here without a user object, so it should be safe to just spread the user (unless typescript does not enjoy it?)

jrosen081 (approved):
Two small things, but it works super well!

frontend/src/pages/Dashboard/data/reducer.ts
cindy1u0: ya i tried doing user: { ...state.user } it gives me an error lol so that's why i added a check there if that's what u meant

frontend/src/firebase/firebaseInteractor.ts
cindy1u0: ya for update im thinking is that for old users which don't have profileicon field in firebase right, then u technically are updating it but if u dont have a field, wouldn't that just give u an error?

frontend/src/pages/Dashboard/data/reducer.ts
jrosen081: I think it might make sense to do a `//@ts-ignore` since I think it would make the code much more readable. Up to you tho

sarah-witt (commented):
Looks so great! One small thing I noticed was that the icons don't all fit on mobile view

frontend/src/pages/Dashboard/data/actions.ts
cindy1u0: good catch lol

cindy1u0: also ya will change styling for mobile

frontend/src/pages/Dashboard/data/reducer.ts
cindy1u0: okie i think ima keep it this way just cuz if state.user is undefined i dont want it to blow up hehe

PR #81, opened 2021-03-23
Show all students

angelinahan:
Added researcher dashboard's view of all student and a dropdown to filter by alphabetical, age, and session
Screen Shot 2021-03-22 at 8 22 40 PM
Screen Shot 2021-03-22 at 8 23 48 PM
Screen Shot 2021-03-22 at 8 24 00 PM


frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -507,4 +507,31 @@ export default class FirebaseInteractor {
...
+        accountType: studentData.accountType,
+        age: studentData.age,
+        sessionId: studentData.sessionId,
+        onAssessment: studentData.onAssessment,
+        currentInterventionOrAssessment:
+          studentData.currentInterventionOrAssessment,
+        daysActive: studentData.daysActive,
+        profileIcon: studentData.profileIcon,
+      };
+    });
```
jrosen081: This is awesome! Can we combine this with the `getCurrentUser` so have a function that takes in an Any and returns a user? This will make all changes to the User object (if there are any) consistent in all places.

frontend/src/pages/Dashboard/ResearcherDashboard.tsx
```diff
@@ -0,0 +1,241 @@
...
+}
+
+const StudentCard: FunctionComponent<StudentCardParams> = ({
+  student,
+}) => {
+  return (
+    <StudentCardContainer>
+      <Avatar
+        src={
+          student.profileIcon
```
jrosen081: You can use `??` to make this simpler, so `student.profileIcon ?? "other URL"`

frontend/src/pages/Dashboard/ResearcherDashboard.tsx
```diff
@@ -0,0 +1,241 @@
...
+  return 0;
+};
+
+const sortByAge = (studentA: User, studentB: User): number => {
+  return studentA.age - studentB.age;
+};
+
+const sortBySession = (studentA: User, studentB: User): number => {
+  if (
+    studentA.sessionId === undefined ||
```
jrosen081: I don't think sessionId will ever be undefined, so you probably do not need this check here.

frontend/src/pages/Dashboard/ResearcherDashboard.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+interface ResearcherDashboardParams {
+  students: User[];
+}
+
+const ResearcherDashboard: FunctionComponent<ResearcherDashboardParams> = ({
+  students,
+}) => {
+  const [sortBy, setSortBy] = useState(() => sortByName);
+  const [sortLabel, setSortLabel] = useState("alphabetical");
```
jrosen081: I don't think we need to store both of these, since one is derived from the other, so it might make sense to only store the label, and use that to decide how to sort?

frontend/src/pages/Dashboard/data/reducer.ts
```diff
@@ -33,6 +33,11 @@ const dashboardReducer = (
         totalWordsLearned: action.payload.totalWordsLearned,
         error: undefined,
       };
+    case ActionTypes.GET_DATA_FOR_RESEARCHERS_SUCCESS:
```
jrosen081, line 36: I think everywhere that we set the rest of the state to be undefined, we should do it for this data as well.

jrosen081 (approved):
This looks super awesome. One small comment is that the log out button and sidebar should take up the whole screen and not scroll. Not sure if that should be part of this pr or not, but definitely something that should be updated at some point!

frontend/src/pages/Dashboard/ResearcherDashboard.tsx
```diff
@@ -0,0 +1,241 @@
...
+    return -1;
+  }
+  return 0;
+};
+
+const sortByAge = (studentA: User, studentB: User): number => {
+  return studentA.age - studentB.age;
+};
+
+const sortBySession = (studentA: User, studentB: User): number => {
```
stefan-philip, line 190: I'm seeing some weird behavior when resizing the screen while sorting by session.

 I don't know why this is happening exactly, but an idea for fix:
- a sort by name as a secondary sort in the case that the students are at the same session



https://user-images.githubusercontent.com/17476012/112082721-378a6180-8b5c-11eb-807d-9677e49d7369.mp4

stefan-philip (commented):
This looks great! I love how responsive it is. And the sorting dropdown is so crispy great job! Left a comment about a behavior I noticed. I don't know if is the same on your browsers, but try to take a look at it :)

frontend/src/pages/Dashboard/ResearcherDashboard.tsx
jrosen081: I do see that, great catch!

frontend/src/firebase/firebaseInteractor.ts
sarah-witt: Do you mind explaining this a little more?

frontend/src/pages/Dashboard/ResearcherDashboard.tsx
sarah-witt: we still have a lot of old user data that don't have a session id so this doesn't sort without the check

frontend/src/pages/Dashboard/ResearcherDashboard.tsx
sarah-witt: good catch! we didn't check if both session ids were undefined. this fixed it for me on chrome, let me know if it works for you!

frontend/src/firebase/firebaseInteractor.ts
jrosen081: Right now we have the same code in `getUser`(I think this one is the actual correct name) and now this function. Since they are the same, can we consolidate into one place (this will also fix the problems of you not having values for sessionId and such, since that code has default values in place)?

jrosen081:
A small thing I noticed is that the sorting changes nicely, but the actual "sort by: ..." dropdown never changes what it says it is sorting by.

frontend/src/firebase/firebaseInteractor.ts
sarah-witt: makes sense, let me know if what I changed looks like what you had in mind!

sarah-witt:
> A small thing I noticed is that the sorting changes nicely, but the actual "sort by: ..." dropdown never changes what it says it is sorting by.

Oh true, that used to change but stopped after I removed the` useState()` for `sortLabel`. We can look into it to see if there's a way to update the it without using `useState()`.

jrosen081:
I think it might make sense to store the sort label with `useState` and use that to determine how the sorting will be done if that makes sense. Like instead of storing the sorting function, compute the sorting function from the label name.

frontend/src/firebase/firebaseInteractor.ts
jrosen081: That looks perfect!

sarah-witt:
> I think it might make sense to store the sort label with `useState` and use that to determine how the sorting will be done if that makes sense. Like instead of storing the sorting function, compute the sorting function from the label name.

Gotcha, updated!

cindy1u0 (approved):
Looks awesome!

jrosen081 (approved):

PR #82, opened 2021-03-23
Have a log out button on the error screen

jrosen081:
I noticed you could get yourself into a situation where the dashboard is broken (like if you have an old account without data) and the error screen only sends you to the dashboard (the first step of recursion is trust the recursive call?). This adds a log out button.


stefan-philip (approved):
Great catch

PR #84, opened 2021-03-25
Side panel menu

stefan-philip:
I tried rotating the dropdown caret on click but I couldn't get it. If anyone knows how to, let me know!


jrosen081:
A couple comments on how this works. If you go from small sized screen with dropdown hidden to full screen, you do not see all of the options. Also, when in the hidden mode, we should show the selected row, not just the top row. And for the rotation, you should be able to use a [transform](https://www.w3schools.com/cssref/css3_pr_transform.asp)

stefan-philip:
omg I can't believe I missed that I'm embarrassed 😳 thanks!!

jrosen081:
Another super small comment. We should probably make the drop down go down when you click on any part of the row, not just the caret

karmenlu:
> Another super small comment. We should probably make the drop down go down when you click on any part of the row, not just the caret

jrosen081:
This works super well. One other super small thing is that when the dropdown is open, I think the log out button should be at the bottom. Right now, you cannot log out on mobile 😬

frontend/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -543,6 +602,191 @@ const SessionCard: FunctionComponent<SessionCardParams> = ({
...
 
+interface MenuButtonPanelProps {
+  isDropdown: boolean;
+  signOutHandler: () => void;
+}
+
+interface MenuDropdownButtonProps {
+  rotationDegrees: number;
+}
+const MenuDropdownButton = styled.img`
```
jrosen081, line 613: Can we rename this to say it is the drop down caret?

jrosen081 (approved):
Looks awesome, just one small comment!

PR #88, opened 2021-03-28
Make sidebar sticky when scrolling

jrosen081:
If you go to the researcher dashboard. The sidebar now is sticky so when you scroll it stays in place.

### DIFF
--- a/frontend/src/pages/Dashboard/Dashboard.tsx
+++ b/frontend/src/pages/Dashboard/Dashboard.tsx
@@ -111,12 +111,15 @@ const DashboardContainer = styled.div`
 
 const MenuContainer = styled.div`
   display: flex;
+  position: sticky;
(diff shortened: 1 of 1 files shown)


stefan-philip:
Very cool! Did not know `position: sticky` existed! One small comment. The side bar's rotated corner is hidden now, you have to scroll up to see it. We probably want some margin on the top

stefan-philip:
> Very cool! Did not know `position: sticky` existed! One small comment. The side bar's rotated corner is hidden now, you have to scroll up to see it. We probably want some margin on the top

This was not something that you introduced btw. The hidden corner is already in the main build, just wanna know if you can fix that

jrosen081:
@stefan-philip Good catch. I updated it!

stefan-philip (approved):

PR #90, opened 2021-03-30
Play audio file

angelinahan:
This allows us to play audio files instead of text to speech prompts. Right now the audio urls are hardcoded but we'll get them from firebase eventually.


frontend/src/components/TriggeredPrompt.tsx
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    setTimeout(() => {
+      stopAudio(prompt1);
+      if (triggerSecondPrompt) {
+        prompt2.play();
+      } else {
+        prompt1.play();
+      }
+    }, 1000);
```
jrosen081: Instead of hardcoding this value here, do we want to use the `duration` of prompt1 and then add like 3000 to make sure that the second prompt starts after the second. Or another idea would be to listen for the end of the first prompt to play the second prompt on a timeout? I think the 1000 might work for some prompts and not others, which is why I am wondering what the right move is.

jrosen081 (approved):
This looks awesome! It would be awesome if you could hook this up to firebase, as most of the fields are already stored, just need a new url!

frontend/src/components/TriggeredPrompt.tsx
stefan-philip: Just did this!

stefan-philip:
**UPDATE:**
We will be doing all of the Firebase related stuff in another PR. For now, we have hardcoded prompts.
This PR is ready for review!

jrosen081 (approved):

PR #93, opened 2021-04-04
Arrange session panels

sarah-witt:
- arrange session panels and add title and redirect button to session page
vocab

### DIFF
--- a/frontend/src/firebase/firebaseInteractor.ts
+++ b/frontend/src/firebase/firebaseInteractor.ts
@@ -468,13 +468,6 @@ export default class FirebaseInteractor {
     userId: string,
     sessionId: number
   ): Promise<SessionStats> {
-    let user = await this.getUser(userId);
-    if (user == null) {
-      throw new Error(
-        `Requested user ${userId} session dashboard does not exist`
-      );
-    }
-
     let interventionForSession = (
       await this.db
         .collection("interventions")
@@ -512,7 +505,6 @@ export default class FirebaseInteractor {
 
     return {
       userId: userId,
-      userName: user.name,
       sessionId: sessionId,
       interventionDuration: interventionDuration,
       assessmentDuration: assessmentDuration,
--- a/frontend/src/firebase/firebaseInteractor.ts
+++ b/frontend/src/firebase/firebaseInteractor.ts
@@ -468,6 +468,13 @@ export default class FirebaseInteractor {
     userId: string,
     sessionId: number
   ): Promise<SessionStats> {
+    let user = await this.getUser(userId);
+    if (user == null) {
+      throw new Error(
+        `Requested user ${userId} session dashboard does not exist`
+      );
+    }
+
     let interventionForSession = (
       await this.db
         .collection("interventions")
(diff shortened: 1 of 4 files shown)


frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -468,6 +468,13 @@ export default class FirebaseInteractor {
     userId: string,
     sessionId: number
   ): Promise<SessionStats> {
+    let user = await this.getUser(userId);
+    if (user == null) {
+      throw new Error(
+        `Requested user ${userId} session dashboard does not exist`
+      );
+    }
```
jrosen081: I don't think we need this here. The dashboard already has the requested user, can we pass it to the SessionDashboard through there?

frontend/src/pages/Dashboard/SessionDashboard.tsx
```diff
@@ -12,8 +13,70 @@ interface StatParams {
...
+const SessionBody = styled.div`
+  display: flex;
+  flex: wrap;
+  @media (max-width: 900px) {
+    flex-direction: column;
+  }
+`;
+
+const WordContainer = styled.div`
+  width: 60%;
```
jrosen081: Can we use flex here, since this is in a container with `display: flex`?

frontend/src/pages/Dashboard/SessionDashboard.tsx
```diff
@@ -12,8 +13,70 @@ interface StatParams {
...
+  font-family: "Rubik";
+  font-size: 56px;
+  font-weight: 700;
+  @media (max-width: 900px) {
+    font-size: 36px;
+  }
+`;
 const StatContainer = styled.ul`
-  padding: 24px;
+  width: 40%;
```
jrosen081: Same here regarding flex.

jrosen081 (approved):
This looks awesome. A few small comments and then 🚀

frontend/src/firebase/firebaseInteractor.ts
sarah-witt: we tried this, but the name would show up as `undefined` if we didn't get to the page through the user's dashboard (like if a researcher were to just visit the link directly rather than click through the user page), since`requestedStudent` would never be set in that case.

jrosen081: Gotcha. I'm not sure if this is better than figuring out why requested student turns up as `undefined`. I am good with it, but do wonder why it does not get fetched correctly :shrug:

sarah-witt: i'm pretty sure it's because we check for `params.id` here https://github.com/sandboxnu/vocab-buddy/blob/master/frontend/src/pages/Dashboard/Dashboard.tsx#L534. I could change it to also load the requested user on a session page there but either way we still have to fetch the user

jrosen081: Yeah I think changing it to use the existing behavior through the dashboard might be better, so that when we go back to the dashboard the data is already loaded. If you could update it that would be awesome.

sarah-witt: updated!

jrosen081 (approved):

PR #94, opened 2021-04-05
See assessment and intervention results

jrosen081:
image


sarah-witt (commented):
Looks great! This is a nit but in the design there is a small purple border around the word groupings, and i think it might be helpful to include it here so the background is more distinct

jrosen081:
@sarah-witt Good catch. The border should be shown now!

sarah-witt (approved):

PR #95, opened 2021-04-06
Fixed up issue regarding to assessment session id and total sessions

jrosen081:
Previously we would set the session id to 0 for all assessments. What this means is that we would never go past session 2. As well, we had one too many sessionId possibilities, now there is the correct amount.
-1 -> Pre-assessment
0-7 -> The 8 intervention/assessment sessions
8 -> Finished


cindy1u0 (approved):
LGTM!!

stefan-philip (approved):
omg

PR #96, opened 2021-04-08
Audio from firebase

angelinahan:
Gets audio from firebase instead of using hardcoded values, and only displays next button after the second prompt has finished playing (for interventions)


frontend/src/components/TriggeredPrompt.tsx
```diff
@@ -72,6 +74,10 @@ const TriggeredPrompt: FunctionComponent<TriggeredPromptProps> = ({
   const prompt1 = new Audio(prompt1Url);
   const prompt2 = new Audio(prompt2Url);
 
+  prompt2.onended = () => {
+    secondPromptFinishedHandler?.();
```
jrosen081, line 78: Do you need the `.` here? Looks kinda weird, but might be needed

frontend/src/components/TriggeredPrompt.tsx
```diff
@@ -93,26 +99,28 @@ const TriggeredPrompt: FunctionComponent<TriggeredPromptProps> = ({
...
-        return () => {
+    if (prompt2Url !== undefined) {
+      getDuration(prompt2, 0, 50, updatePrompt2);
+      if (canPlayPrompt1 && canPlayPrompt2) {
+        // Activities 1 and 4 do not set triggerSecondPrompt, but do set promptDelay
+        if (promptDelay !== undefined) {
+          playPrompt1();
+          setTimeout(() => {
+            playPrompt2();
+            triggerSecondPrompt = true;
```
jrosen081: What does this do? I think on rerender this will be ignored, so will this work without changing this value?

jrosen081 (approved):
Looking awesome just a few small comments.

frontend/src/components/TriggeredPrompt.tsx
stefan-philip: Agreed it looks a little odd, but it is needed

frontend/src/components/TriggeredPrompt.tsx
stefan-philip: This is set so that the replay button replays the audio from prompt2 instead of prompt1, after prompt2 has actually been played.

stefan-philip: Jkjk I'm bugging, I think you're right. lemme do some testing

PR #98, opened 2021-04-11
Don't allow students to click on session stats

sarah-witt:
- right now you can click on the session but it just leads to the error page


jrosen081 (approved):
Good catch!

PR #99, opened 2021-04-11
Added the download option to the student dashboard

jrosen081:
This creates a zip file with all 8 session's data in csv format.


sarah-witt (approved):
works well for me!

PR #102, opened 2021-04-21
Update user settings

stefan-philip:
This allows users to update their name, age, email, and password.

On success, we show a success toast.
On error, we show an error toast with either an error message from Firebase, or a validation error.
Screen Shot 2021-04-20 at 9 28 02 PM

Screen Shot 2021-04-20 at 9 28 28 PM

Screen Shot 2021-04-20 at 9 29 08 PM


frontend/src/components/Toast.tsx
```diff
@@ -0,0 +1,50 @@
...
+  }
+  margin: auto;
+  z-index: 1000;
+`;
+
+interface ToastProps {
+  message?: string;
+  onClose: () => void;
+  topMargin?: string;
+  alertType?: "success" | "info" | "warning" | "error" | undefined;
```
jrosen081: I don't think this needs the `| undefined`, since the `?` after `alertType` makes this allowed to be undefined.

frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -624,6 +625,59 @@ export default class FirebaseInteractor {
...
+      await this.updateCurrentUser({ name: newName });
+    }
+    if (newAge !== undefined) {
+      await this.updateCurrentUser({ age: newAge });
+    }
+    return this.getUser(undefined);
+  }
+
+  async reauthenticateUser(password: string) {
+    const user = firebase.auth().currentUser;
```
jrosen081: I think we have `this.auth` as `firebase.auth()`, so we may want to use that.

frontend/src/pages/Dashboard/Settings.tsx
```diff
@@ -0,0 +1,266 @@
...
+  };
+
+  useEffect(() => {
+    if (networkErrorShown) {
+      setShowSuccessToast(true);
+    }
+  }, [user]);
+
+  return (
+    <>
```
jrosen081: Do we need the wrapper `<></>`? It might be necessary, but hard to tell from the git diff.

frontend/src/pages/Dashboard/data/actions.ts
```diff
@@ -108,6 +113,35 @@ export const ChangeProfileIcon = {
   }),
 };
 
+interface UpdateUserSettingsSuccessResponse {
+  user?: User;
```
jrosen081, line 117: Could this possibly not have a user? I wonder if this should be non-optional?

jrosen081 (approved):
This looks perfect! I saw one or two small things that need to be updated, but other than these good.

1. The error modal does not show up at the top, but rather the bottom, of the login page (and I assume the create user page as well)
2. We don't have the review words option, so can you remove it? I know it wasn't part of the ticket, but it could be awesome if you do that since relates to the switching of dashboard.

frontend/src/pages/Dashboard/data/actions.ts
stefan-philip: It should always return the current user, provided that no exceptions are thrown. But when I make it non-optional, we get errors in the saga saying that the field may have not been initialized from the call to firebaseInteractor

jrosen081: Ok works for me!

PR #113, opened 2022-08-12
modified blocker as per requested changes

### LINKED ISSUE #107: "Are you still there?" notification changes
**Problem:** The time for the 'are you still there' notification is too short and its prompt needs changing
**Solution:** Make the time longer! And change the prompt
**Description:**
Changes will likely occur in [this file here](https://github.com/sandboxnu/vocab-buddy/blob/master/frontend/src/components/Blocker.tsx#L135). The researchers also want a button to be clicked as the prompt instead of a star being dragged across a line as a way for students to show they're still on the page. 

**Specs:**
- [x] Lengthen wait time - add 30 seconds to whatever the time originally is 
- [x] Change prompt to button instead of dragging across a star

See [this file](https://www.notion.so/sandboxnu/Vocab-Buddy-Reported-Bugs-17922cac823041c189795da750e2b2fe) for more information (thank you Cindy u r literally amazing)

clue4:
closes #107 

*** at first I was a bit confused with the wording for the time additions -- I'd first thought it meant that the notification disappears after too short a time, but that's def not the case since it doesn't close unless you manually make it close (but will also check with researchers about this soon ;-; :D). So, I've made it that it takes an additional 30 seconds to show up on whichever pages use it. 

anyway changes were p quick! shouldn't be too hard to review


dillydally414 (approved):
Want a break from the ads? If you tap now to watch a short video, you'll receive 30 extra seconds of ad-free vocab buddy

PR #116, opened 2023-07-30
audio echo fix

### LINKED ISSUE #108: Audio Echo issues
**Problem:** Some of the audio recordings have an echo to them during the activities 
**Solution:** Make the audio echo disappear
**Description:**
See [this video here](https://watch.screencastify.com/v/G9WAZjt16xLmzZ44cdPb) for a recording of the audio echo (starting at 11 seconds). The audio recordings themselves are fine / should be fine. Cindy noted that it might be a browser issue and that it doesn't seem to happen on firefox. The researchers said that this echo issue seems to be happening on most of the items on the initial assessment that students complete in the beginning of the study; in another case every other audio prompt had echoes. 

**Specs:**
- [ ] Investigate the audio issue / test different browsers
- [ ] Fix issue (idk what else to say honestly) 

**Resources** See [this page](https://www.notion.so/sandboxnu/Vocab-Buddy-Reported-Bugs-17922cac823041c189795da750e2b2fe) for additional information + potential ideas.

dillydally414:
Closes #108 (and another bug where we let people hit next without selecting an image)


frontend/src/components/TriggeredPrompt.tsx
```diff
@@ -56,113 +29,101 @@ const TriggeredPrompt: FunctionComponent<TriggeredPromptProps> = ({
...
-      const interval = setInterval(
-        playPrompt1,
-        prompt1Duration * 1000 + 8000
-      );
-      return () => {
-        stopAudio(prompt1);
-        stopAudio(prompt2);
-        clearInterval(interval);
-      };
+  const [audio1, setAudio1] = useState(new Audio(prompt1Url));
```
clue4, line 27: in the code I'm getting that the variables promptDelay and currentWord are never used  by TriggeredPrompt -- might as well just remove it then if that's still the case

frontend/src/components/TriggeredPrompt.tsx
```diff
@@ -56,113 +29,101 @@ const TriggeredPrompt: FunctionComponent<TriggeredPromptProps> = ({
...
+    ) {
+      audio1.currentTime = 0;
+      audio2.currentTime = 0;
+      audio1.pause();
+      audio2.play();
+      setLastPlayed("audio2");
+    } else if (audio1.paused && audio2.paused) {
+      audio1.currentTime = 0;
+      audio1.play();
+      setLastPlayed("audio1");
```
clue4, line 58: so just to clarify with the logic of this if statement, if neither of these cases are true, that means an audio is already playing, so we don't want to do anything?

also I've beeen seeing the triggerSecondPrompt variable -- just curious if there's an equivalent for 'triggering' the first prompt, or is that just the default/not necessary?

frontend/src/components/TriggeredPrompt.tsx
```diff
@@ -3,6 +3,9 @@ import React, {
   ReactElement,
   useEffect,
   useState,
+  useMemo,
```
clue4: useMemo is never used -- ok to remove then?

frontend/src/components/TriggeredPrompt.tsx
```diff
@@ -56,113 +29,101 @@ const TriggeredPrompt: FunctionComponent<TriggeredPromptProps> = ({
...
+    "audio1"
+  );
+  const [
+    prevTriggerSecondPrompt,
+    setPrevTriggerSecondPrompt,
+  ] = useState(triggerSecondPrompt);
+  const [prevPrompt1Url, setPrevPrompt1Url] = useState(prompt1Url);
+  const [prevPrompt2Url, setPrevPrompt2Url] = useState(prompt2Url);
+
+  const [audio2, setAudio2] = useState(new Audio(prompt2Url));
```
clue4: this is literally the smallest of nitpicks so honestly please feel free to ignore, but could we move this line to below line 32 so that they're next to each other oop

clue4 (approved):

clue4:
good stuff + nice cleanup

frontend/src/components/TriggeredPrompt.tsx
dillydally414: ugh i'm so used to eslint handling everything for me 😩

frontend/src/components/TriggeredPrompt.tsx
dillydally414: > so just to clarify with the logic of this if statement, if neither of these cases are true, that means an audio is already playing, so we don't want to do anything?

yes, this way the prompt won't be interrupted if it's already playing

> also I've beeen seeing the triggerSecondPrompt variable -- just curious if there's an equivalent for 'triggering' the first prompt, or is that just the default/not necessary?

first prompt is automatically played on page load, whereas the second prompt is only triggered in certain cases (for some activities it will play as soon as audio1 is done, while for some a button needs to be pressed first)

dillydally414: also in assessments there is no second prompt

PR #117, opened 2023-07-30
remove correct/incorrect from image names

### LINKED ISSUE #106: Images give away answers when downloaded
**Problem:** In the assessments and activities, students can download images and obtain answers via file name - the images have "correct" or "incorrect" in their file name
**Solution:** Change the names of the images and files so that the answer is not given away. (If this doesn't work, potential other solutions include somehow disabling right click)
**Description:**
We need to make it so that even if a student downloads an image off of their activity, they still cannot tell what the answer is. Images are uploaded to our codebase via a script. 

You can view our firebase (go to firestore database) to see the how the words are stored and how the images associated with each word give away the answer. 

**Specs:**
- [ ] Find out where the images are being named as they're being uploaded (one place might be [this script here](https://github.com/sandboxnu/vocab-buddy/blob/master/uploadscript/script.js#L47))
- [ ] Change file names to something that is not as obvious to avoid giving away the answer as they're being uploaded

**If blocked:** #engineering-support, @cindy1u0 
See [this notion page](https://www.notion.so/sandboxnu/Vocab-Buddy-Reported-Bugs-17922cac823041c189795da750e2b2fe) for more information.

dillydally414:
Closes #106

### DIFF
--- a/uploadscript/script.js
+++ b/uploadscript/script.js
@@ -42,14 +42,15 @@ const performUpload = async () => {
         let incorrectFolder = wordFolder + "incorrect/";
         let wordRef = db.collection("words").doc();
         // Create the base assessment information
+        const ids = [1, 2, 3, 4].sort((_a, _b) => Math.random() - 0.5)
         let wordBaseInfo = {
             value: word,
-            correctImage: await uploadFileToFirebaseStorage(correctFolder + "correct_assessment.jpg", wordRef.id + "/correct_assessment.jpg"),
+            correctImage: await uploadFileToFirebaseStorage(correctFolder + "correct_assessment.jpg", wordRef.id + `/assessment${ids[0]}.jpg`),
             dateCreated: new Date(),
             incorrectImages: [
-                await uploadFileToFirebaseStorage(incorrectFolder + "incorrect_assessment1.jpg", wordRef.id + "/incorrect_assessment1.jpg"),
-                await uploadFileToFirebaseStorage(incorrectFolder + "incorrect_assessment2.jpg", wordRef.id + "/incorrect_assessment2.jpg"),
-                await uploadFileToFirebaseStorage(incorrectFolder + "incorrect_assessment3.jpg", wordRef.id + "/incorrect_assessment3.jpg"),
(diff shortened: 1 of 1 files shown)


uploadscript/script.js
```diff
@@ -42,14 +42,15 @@ const performUpload = async () => {
         let incorrectFolder = wordFolder + "incorrect/";
         let wordRef = db.collection("words").doc();
         // Create the base assessment information
+        const ids = [1, 2, 3, 4].sort((_a, _b) => Math.random() - 0.5)
         let wordBaseInfo = {
             value: word,
-            correctImage: await uploadFileToFirebaseStorage(correctFolder + "correct_assessment.jpg", wordRef.id + "/correct_assessment.jpg"),
+            correctImage: await uploadFileToFirebaseStorage(correctFolder + "correct_assessment.jpg", wordRef.id + `/assessment${ids[0]}.jpg`),
```
clue4, line 48: does the first part with the correctFolder stuff also need to be changed? or is that internal / just organizing them for the researchers

clue4 (commented):
still having the issue where I can download an image (or control + open in a new tab) and see "correct_assessment" or "incorrect_assessment" automatically in the name:

url example: https://storage.googleapis.com/vocab-buddy-53eca.appspot.com/6XrSF9PqycgDIZF93KSM/correct_assessment.jpg

download example -- this name was the default answer when I clicked  "save image as"
image
  
I see the file you changed was the upload script file -- maybe we need to remove all the images from firebase and then upload them again?

uploadscript/script.js
dillydally414: i believe this is coming from the local filesystem so retaining the existing naming convention allows us to keep distinctions between correct + incorrect, but once we upload them they will lose that distinction (by name at least)

dillydally414:
@clue4 yeah I will need to re-upload the words, I confirmed this works with the example myword and otherword that I had locally but I don't have access to the actual folder of words. Is this something that you have access to or do I need to ask the researchers?

clue4 (approved):
approved after reading responses

+ answer to your q I put in slack lol

dillydally414:
Going to wait to merge this until I actually perform the fixing upload so that I don't forget

dillydally414:
i lied i actually am gonna merge this cause i think there might be conflicts

PR #119, opened 2023-07-31
Assessments bug

### LINKED ISSUE #109: Incorrect words show up in assessments
**Problem**: Not all words a student is tested for in an intervention show up in their assessments later, or incorrect words show up
**Solution**: Make sure the correct and applicable words show up during assessments after their respective interventions
**Description**: 
Here is an example the researchers provided:
- **Intervention Session 1**: the words drenched, plump, and consume are taught.
- **Session 1 Follow-up Assessment**:  drenched, plump, and consume are assessed.
   - The student must complete the Session 1 Assessment before they are able to move on to Session 2. 
- **Intervention Session 2:** alert, collide, and delicate are taught (3 new words).
- **Session 2 Follow-up Assessment**: alert, collide, and delicate are assessed.
   - The student must complete the Session 2 Assessment before they are able to move on to Session 3. 

However, there are cases where only one word instead of all three words is assessed. There are also cases where after moving on to a new session, previous session words get tested, as per the researchers:
- Some of the in-between/follow-up session assessments only assess one of the words from the previous session instead of assessing all 3 words. This happened on the Session 3 assessment for me. I answered one question for Session 3 (for one word), then the balloons prompt came up, and then I was brought back to the dashboard. I was not tested on the other 2 words from Session 3. I was then able to select the next session “Session 4” and complete the first activity for Session 4… when I clicked “next” it brought me to an assessment item for Session 3 for the other 2 words. It seems that the assessments are out-of-order/mixed up for some of the sessions. Since every student will have a different set of words they are taught/assessed on, it's difficult to identify every instance of this, but it did happen several times during testing. 

**Specs**:
- [ ] Investigate issue -- see if there's any consistency between what causes this
- [ ] All words in a session should be tested by the follow up assessment, not just one word
- [ ] Make sure assessments are not out of order  
   - [ ] if you go onto a new session, words from previous session should not be assessed in the next test, vice versa 

**Resources**: [this lovely notion doc](https://www.notion.so/sandboxnu/Vocab-Buddy-Reported-Bugs-17922cac823041c189795da750e2b2fe) here again for potential ideas/solutions <3

dillydally414:
Closes #109 I think maybe we'll see


frontend/src/pages/Interventions/ActivitiesComponent.tsx
```diff
@@ -19,6 +23,21 @@ const ActivityComponent: FunctionComponent<ActivityComponentProp> = ({
   activities,
   updateIntervention,
 }): ReactElement => {
+  const activity2ImageUrls = useMemo(() => {
```
dillydally414, line 26: unrelated bug, was just getting annoyed cause this kept re-rendering and then it would sometimes show a fake error for a split second

frontend/src/pages/Assessments/Quiz.tsx
```diff
@@ -57,8 +57,8 @@ const Quiz = ({
   }
 
   useEffect(() => {
-    if (!assessment) getAssessment(params.id);
-  }, [assessment, getAssessment, params]);
+    if (!assessment && !isFinished) getAssessment(params.id);
```
dillydally414, line 60: i think this was the root of the problem, right after assessments would be submitted there was a chance that this useEffect would re-run because the assessment would be set to undefined, but then getAssessment would reload in the same assessment - meaning that when the user would come back for new assessment it would read in the old assessment. now we don't reload if the isFinished flag is true. If this doesn't fix it my other potential fix is to cross-check the `params.id` with the `assessmentId` cause we don't want those to be out of sync

clue4: dillon you're a legend
i'll look at this tomorrow (but lowkey any better ideas of testing besides "just run through assessments and stuff a lot"

frontend/src/pages/Interventions/Activities.tsx
```diff
@@ -122,12 +122,9 @@ const Activities: FunctionComponent<ActivityProps> = ({
           let durationInSeconds =
             (curDate.getTime() - activityStartTime.getTime()) / 1000;
           setActivityStartTime(curDate);
-          if (
-            currentWordIdx === wordList.length - 1 &&
-            nextActivityIdx === 0
-          ) {
-            history.push(`/interventions/reward`);
+          if (nextWordIdx === 0 && nextActivityIdx === 0) {
```
clue4, line 108: what's the reasoning behind this change? (not saying it's wrong, just curious

clue4 (commented):
gonna be so real i was not in the mood to do 20398409238 assessments and interventions to see if that bug would pop up, but I will do that tomorrow 

code looks good, will do that check tomorrow and then approve / add more comments as follows

frontend/src/pages/Interventions/Activities.tsx
dillydally414: wanted to be consistent lol, i think it's basically the same calculation but since we're using nextActivityIndex i wanted to use nextWordIndex for clarity. if you're referring to the line swap below this it's because i initially thought the finishedIntervention call was being interrupted by the history.push call (it wasn't)

clue4: ahh i see

also I said I was gonna test this today, but the api key expired? so can't really do much

dillydally414:
I tested and I _think_ I fixed this 🤔  gonna merge ahead of the client meeting tomorrow so they can test as well

clue4 (approved):
didn't run into the assessments issue, so it seems good?

one thing I did see, not sure if it's just the deployment -- fonts seem to have changed / it's now in times new roman or some similar serif font. iirc, fonts used to not be serif -- could just be an issue with loading tho
image

PR #118, opened 2023-07-30
include more info in downloads/exports and support all-student download

dillydally414:
Closes #112 
Closes #115

PR #120, opened 2023-08-02
Database Mods

dillydally414:
- Reuploaded words so correct/incorrect is now removed as per #106 (this needs to be verified on the deploy)
- Wiped database (sorry not sorry)
- Changed database structure to have overall less documents, preferring nested objects over sub-collections
  - This will hopefully reduce the number of reads for downloads 🤞 still need to test this theory.... in the meantime it's gonna break the other PRs LOL

Changes from #118 are also included in this since it's branched off of `download-improvements`, so this pr deploy hopefully should sort of replace the pr deploy for that pr

#119 is going to be put on the backburner until this is complete so we can see if it resolves those issues

One more change that I might include in this (from the researchers):
> Q: If the student ends with less than 24 incorrect answers, what should they see in their following interventions? 
> A:  For our study we are mostly interested in collecting data for kids who answer 24 items incorrectly in the pre-assessment. For students who answer fewer than 24 items incorrectly, either of these options would be OK:
Option 1: The student completes intervention/assessment activities only for the words that they answered incorrectly on the pre-assessment (i.e., if they answered 12 items incorrectly, they would complete intervention activities for those 12 words and then the program would end). 
Option 2: If Option 1 isn't possible, the program can end after the pre-assessment is completed for students who answered <24 incorrect items.

Probably going to try out option 1 where we just only populate the appropriate number of interventions and then the user ends after their final intervention, but also seems like the issue isn't super important _as long as we handle these cases in some manner without tanking the app_

### DIFF
--- a/frontend/src/firebase/firebaseInteractor.ts
+++ b/frontend/src/firebase/firebaseInteractor.ts
@@ -22,6 +22,7 @@ import {
   WordMapping,
   WordResult,
 } from "../models/types";
+import _ from "lodash";
 
 // Your web app's Firebase configuration
 var firebaseConfig = {
@@ -405,13 +406,10 @@ export default class FirebaseInteractor {
       .filter(([_id, result]) => !result.correct)
       .map(([id, _result]) => id);
     shuffle(incorrectWords);
-    for (let i = 0; i < 8; i++) {
+    const interventionWords = _.chunk(incorrectWords, 3).slice(0, 8); // just in case we have more than 24 for some reason
+    for (let session in interventionWords) {
       // Add 3 words per intervention (wrapping around for now)
-      let wordList = [
-        incorrectWords[(i * 3) % incorrectWords.length],
-        incorrectWords[(i * 3 + 1) % incorrectWords.length],
-        incorrectWords[(i * 3 + 2) % incorrectWords.length],
-      ];
+      let wordList = interventionWords[session];
       let newIntervention = await this.db.collection("interventions").add({
         durationsInSeconds: 0,
         activityIdx: 0,
@@ -420,10 +418,10 @@ export default class FirebaseInteractor {
         // Assign it to the current user
         userId: this.auth.currentUser?.uid,
         // Decide which session the intervention is in
-        session: i,
+        session: Number.parseInt(session, 10),
         results: {},
       });
-      if (i === 0) {
+      if (session === "0") {
         await this.updateCurrentUser({
           sessionId: 0,
           onAssessment: false,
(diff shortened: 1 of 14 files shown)


scripts/upload-words.js
```diff
@@ -21,61 +21,88 @@ let repositoryStartingFolder = "./words";
 
 let allWords = fs.readdirSync(repositoryStartingFolder);
 
+const findFileLocation = (fileLocation) => {
+  if (fs.existsSync(fileLocation)) {
+    return fileLocation;
+  }
```
clue4, line 27: if we're returning the fileLocation anyway if none of the below cases are met, can we get rid of this if statement?

frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -527,53 +525,46 @@ export default class FirebaseInteractor {
...
       | undefined = undefined;
 
     if (assessmentForSession !== undefined) {
       assessmentDuration =
         Math.round(assessmentForSession.data().durationsInSeconds * 100) / 100;
-      let assessmentResults = await assessmentForSession.ref
-        .collection("results")
-        .get();
+      let assessmentResults = assessmentForSession.data()
+        .results as FirebaseAssessmentResults;
```
clue4, line 533: what happens without the cast here /  in other words? does ts complain ;-;

frontend/src/firebase/firebaseInteractor.ts
```diff
@@ -457,23 +461,16 @@ export default class FirebaseInteractor {
...
-        await wordRef.doc("activity3-part3").get()
-      ).data() as Context;
+      let {
+        activity1,
+        activity2,
+        activity3,
+        activity3Part2,
+        activity3Part3,
+        activity4,
+      } = (await this.db.collection("words").doc(word).get()).data()
```
clue4, line 469: yeah this looks a lot nicer than before -- nice cleanup here and everywhere else :o

clue4 (approved):
thank you dillon for all the hard work you've done!

scripts/upload-words.js
dillydally414: I guess technically, but I like this better because we're explicitly saying "if the unaltered file exists we don't need to look through these edge cases" and then the catch all is just "if the file name is really messed up we'll just give up" which will produce a theoretically helpful error message in the console.

frontend/src/firebase/firebaseInteractor.ts
dillydally414: .data() returns `any` since we don't have any knowledge of what data types are in the database from within the code. You can think of this more as an annotation than a cast, so that I can use type hints to see what fields are present on assessmentResults

